PR #7, opened 2017-04-04
update: package.json and README ready for npm

FezVrasta:
To publish the library:

1. Register to https://npmjs.org
2. `npm login` and insert your npmjs data
3. `cd` to Tippy.js folder
4. `npm publish` will publish `typpy.js` into the npm registry

From now on `npm install typpy.js` will install your library.
To publish a new version, bump the version number in `package.json` and run `npm publish` again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,9 @@
 Tippy.js is a lightweight, pure JS tooltip library powered by Popper.js.
 
 View the documentation and demo here: https://atomiks.github.io/tippyjs/
+
+## Installation
+
+```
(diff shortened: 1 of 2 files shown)


atomiks:
I've published it, can you check if it's working properly? 😛

FezVrasta:
It works, you may want to publish to npm a transpiled version to avoid consumers the need of a transpilation step.

For instance, with Popper.js I provide a `.es6.js` version which is transpiled to ES6, and a `.js` version that is compatible with ES5.

PR #38, opened 2017-04-29
Add event parameter for wait callback

ismail-codar:
Sometimes you may need to check with event.target before calling show
callback.


atomiks:
Seems like this is a really "old" (development has been really quick because this lib is new 🤣 ) version, I can add it in manually to v0.11

atomiks:
This is in v0.11.2 now

PR #83, opened 2017-07-28
Add support for array of elements as selector parameter

artyomtrityak:
For apps it can be helpful to pass array of elements instead of doing it one by one.
For example React.js app

```jsx
const titleEls =  Array.prototype.slice.call(
    ReactDOM.findDOMNode(this).querySelectorAll('[title]')
);
tippy(titleEls, {arrow: true});
```

P.S. Is there any instruction how to run tests? 

```bash
npm run tests
> rollup -c scripts/rollup.tests.config.js
⚠️   No format option was supplied – defaulting to 'es'
https://github.com/rollup/rollup/wiki/JavaScript-API#format
```

P.S2. rollup is not in devDependencies so it make sense to add it

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -196,8 +196,9 @@ tippy.css</code></pre>
               <button class="btn tippy" title="I'm a tooltip!">Text</button>
             </div>
 
-            <p>You can also directly pass in a DOM element:</p>
+            <p>You can also directly pass in a DOM element or array of DOM elements:</p>
 <pre><code class="language-javascript">tippy(document.querySelector('#myElement'))</code></pre>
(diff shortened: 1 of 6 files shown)


atomiks:
Thanks, is there a reason not to use `Array.isArray()` to check if it's an array in this case?

artyomtrityak:
I was thinking about it but `Array.isArray` browser support is IE9+. If you are okay about dropping support IE6-8 i can make those change.

atomiks:
Support has been IE10+ since the beginning

artyomtrityak:
cool, will update PR

atomiks:
I've added your changes myself since there is couple problems with the pull, I'll be releasing it in next ver 1.2 soon

PR #87, opened 2017-08-16
Do not display tooltip if parent element does not exist anymore

artyomtrityak:
For use-case when duration is more than 0 it is possible that element triggered Tooltip has already been removed from the DOM. For this use case currently Tooltip displays at `top: 0; left: 0` position and sticks there unless user clicks anywhere.

This fix checks does parent element exist in DOM tree when it should be displayed.


atomiks:
Only the test file has been changed?

artyomtrityak:
oh, that's unexpected, sorry. Did you submit another file. 
There are some tests which should fail right now, i'll fix them and update review if you're okay about this change.

atomiks:
Yup, it looks good 👍

artyomtrityak:
Cool, will update PR with tests changes

artyomtrityak:
Can you please describe you do you run tests?

I run `npm run tests` and get error

```sh
$ npm run tests

> tippy.js@1.2.0 tests ***/tippyjs
> rollup -c scripts/rollup.tests.config.js


./tests/spec/functional.js → ./tests/functional.js...
[!] Error: You must supply an output format
https://github.com/rollup/rollup/wiki/JavaScript-API#format
Error: You must supply an output format
```

artyomtrityak:
screenshot 2017-08-18 12 42 00

atomiks:
Is there a reason the whole instance is destroyed (instead of just the single) if only that popper's reference element has been removed from the DOM?

artyomtrityak:
@atomiks You're right, i can use destroy method instead. + has to change test for `destory()` instead of `destroyAll()`

Issue #79, opened 2017-07-16
Lots of tooltip with infinite scroll


obsidienne:
Hi,

My website is a song database. Each songs have 3 types of opinion (up / like / down). With that system, our DJs can give their opinions. When a use click give it's opinion, i update the title.
The website uses turbolinks to ensure a continued audio play. So my tippy instance is not destroyed during navigation.

My problem are:
- auto update of title (but they are already a bug on that)
- append to an existing Tippy instance, the new songs added by the infinite scroll. I can destroy but not add a tooltip to an existing instance.

I would like to do : tip.add("CSS selector")

Is there a good way of doing that ?

thanks

atomiks:
Is there a reason why you can't just make a new instance with the same settings?

obsidienne:
Hi,
Maybe, i'm misusing the lib.

Turbolinks implies that every JS objects stay alive accros navigation
turbolinks:load event is the equivalent of the load event. And i use it to init tippy with a ```new Tippy(".tippy")```. 

I use tippy to show the users name liking/disliking the song.
When the user gives is opinions (like/dislike), I do an Ajax call and change the title, when the user scroll I add new songs and need to init Tippy for those new songs and those new title.
I use the destroyAll() but, i have the previous tooltip still displaying with the new tooltip.

obsidienne:
I'm not sure 100% but I think the problem lies in this line of Turbolinks doc

> When possible, avoid using the turbolinks:load event to add event listeners directly to elements on the page body. Instead, consider using event delegation to register event listeners once on document or window.

CreateTrigger in Tippy.js does not use event delegation. Maybe it's the problem.

atomiks:
You can just do `tippy('.tippy')` again and it won't have an effect on existing ones since the `title` attribute gets removed. Even better, `tippy('.tippy[title]')`

I've used it briefly on a project with infinite scroll and did the above and didn't have any problems.

obsidienne:
Yes, sure. ```tippy('.tippy[title])``` is better. I have changed it. 

But the difficulty is more on turbolinks. When turbolinks do a visit (follow a link or restore a previous page). It changes the content of the body. It does not replace it. 
So when the user navigate the site, new tippy() stays in memory so does its appendTo reference. Which is the previous <body>

obsidienne:
Ok, using ```new tippy('.tippy[title]',{appendTo: document.body});```

The function init() is called one time (in case of turbolinks) but the body changes each time and the reference not. 

[...]

PR #95, opened 2017-09-03
Enables passing of function to `appendTo`

padi:
Instead of assigning an Element to `appendTo`,
one can now pass a function that evaluates to an Element

The appendTo function is now evaluated during createToolTips/initialization
of the tippy instance, which is good enough for not leaking appendTo
when reinitializing for various DOM changes e.g. in between
multiple `turbolinks:load` transitions.

This also changes the Defaults.appendTo into a function reevaluating the ‘current’ document.body

### DIFF
--- a/src/js/core/createTooltips.js
+++ b/src/js/core/createTooltips.js
@@ -24,6 +24,12 @@ export default function createTooltips(els) {
     // animateFill is disabled if an arrow is true
     if (settings.arrow) settings.animateFill = false
 
+    // reassign appendTo into the result of evaluating appendTo
+    // if it's set as a function instead of Element
+    if (settings.appendTo && typeof settings.appendTo === 'function') {
+      settings.appendTo = settings.appendTo();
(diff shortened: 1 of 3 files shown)


padi:
I tried converting the appendTo option to always wrap itself into a function whenever an Element object is set, but I found it too difficult to do given that I'm not well acquainted with ES6 and the state of the art javascript tools for debugging.

For example, while I was able to figure out how to run all tests, I don't know how to run the tests individually. I didn't know how to resolve the test failures except by read the backtrace, which makes my feedback loop pretty slow. I had to settle for a solution that I can come up with given the time and energy constraint I've set for myself. XD

This should resolve #79

padi:
@artyomtrityak if you have any suggestions, hit me up on this issue. I was able to run the karma tests via `yarn test` added the test and just made it pass.

atomiks:
Didn't even see this lol, merging now

PR #102, opened 2017-09-22
Added option to override reference object

hmtinc:
Implemented an option to pass in a optional custom reference object so the tooltip can be positioned anywhere on a page and not just bound to a DOM element. 

This should help with implementing tippy in applications where the actual object the tip should appear over isn't a DOM element

Sample Usage : 
```
//Reference object allows for custom positioning
  var refObject = {
    getBoundingClientRect: function (data) {
      return {
        top:  5 + window.pageYOffset,
        left: 5 + window.pageXOffset,
        width: 1,
        height: 1,
    };
    },
    get clientWidth() {
      return 30;
    },
    get clientHeight() {
      return 30;
    },
  };

  var tip = tippy('#object', {}, refObject);
```

The element id which in the example is #object, refers to any DOM element and mainly just serves as a basis to create the popper element.  The reference object actually provides the position .


maxkfranz:
@hmtinc Would you please remove the `dist/*` files from the PR?  Usually those are updated in a repo only when making a new release.

maxkfranz:
@hmtinc It should be possible to just make additional commits on your fork in the master branch to update this PR rather than making a new one, if you like.

PR #103, opened 2017-09-22
Added option to override reference object

hmtinc:
Implemented an option to pass in a optional custom reference object so the tooltip can be positioned anywhere on a page and not just bound to a DOM element.

This should help with implementing tippy in applications where the actual object the tip should appear over isn't a DOM element

Sample Usage :
```js
//Reference object allows for custom positioning
  var refObject = {
    getBoundingClientRect: function (data) {
      return {
        top:  5 + window.pageYOffset,
        left: 5 + window.pageXOffset,
        width: 1,
        height: 1,
    };
    },
    get clientWidth() {
      return 30;
    },
    get clientHeight() {
      return 30;
    },
  };

var tip = tippy('#object', {}, refObject);
```
The element id which in the example is #object, refers to any DOM element and mainly just serves as a basis to create the popper element. The reference object actually provides the position

Edit : Restored distribution files


FezVrasta:
Wouldn't make sense to have Tippy.js accept `refObject` as first argument?

hmtinc:
Are you suggesting that it should be 
```js
var tip = tippy(refObject, '#object', {});
``` 
 I guess that would be a bit more intuitive. I added the refObject to allow for tippy.js to be used with Cytoscape.js where the reference object position doesn't really matter as the Cytoscape.js wraps its anyway. But I guess for general use its better to after it first to avoid having to include an options object.

FezVrasta:
I mean, without '#object' at all. If the selector (first argument) is an object, it will implicity use it as reference object.

This could even allow the API to accept a DOM element directly.

```
tippy(document.querySelector('.foo'), options);
```

hmtinc:
That seems like it could work if you define a html field in options to determine what should be displayed.  Although the main goal of this addition was just to provide a quick way to reposition regular tippy tooltips. 

What I was thinking of was something like this

`<div id="object" title="stuff"></div>`
`tippy("#object", {}, refObject`

This way tippy js can be used as normal where the title parameter still defines the contents of the tooltip, but the position can be redefined to allow for it to appear on top of a image or HTML canvas.

maxkfranz:
I agree with @FezVrasta in that it's better to just have one arg and do type detection.

You can use this to determine type: 

- `typeof arg0 === typeof ''` : string
- `arg0 instanceof HTMLElement` : [HTML element object](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement)
- last else `typeof arg0 === typeof {}` : plain object / ref obj

maxkfranz:
Basically you always need two things for Tippy: (1) target dom ele, (2) the contents dom ele

@hmtinc is saying that the ref object just replaces (1), which makes sense to me.

maxkfranz:
But @FezVrasta is right that (1) and (2) should be able to be specified as `HTMLELement`s.  That's more useful than a selector string imo.

maxkfranz:
Basically, it should look like this:

```js
let refObj = node.popperRef();

let optionsObj = {
  html: document.createElement('div') // whatever div for content
  // maybe some other options...
};

let tippy = new Tippy( refObj, optionsObj );
```

src/js/core/getArrayOfElements.js
```diff
@@ -1,13 +1,17 @@
...
-* @param {String|Element|Element[]} selector
+* @param {String|Element|Element[]|Object} selector
 * @return {Element[]}
 */
 export default function getArrayOfElements(selector) {
   if (selector instanceof Element) {
     return [selector]
   }
 
+  if (selector instanceof Object) {
```
maxkfranz: You need to also check `selector.constructor === Object` if you want to check for plain objects, because `Element` inherits from `Object`.

src/js/core/createTooltips.js
```diff
@@ -16,27 +16,30 @@ let idCounter = 1
 * @return {Object[]} Array of ref data objects
 */
 export default function createTooltips(els) {
+
   return els.reduce((a, el) => {
     const id = idCounter
 
+    console.log(el);
```
maxkfranz: Should remove logs from PR.

dist/tippy.css
```diff
@@ -1 +1 @@
...
...tooltip.light-theme{color:#26323d;box-shadow:0 4px 20px 4px rgba(0,20,60,.1),0 4px 80px -8px rgba(0,20,60,.2);background-color:#fff}.tippy-popper .tippy-tooltip.light-theme[data-animatefill]{background-color:transparent}.tippy-tooltip{position:relative;color:#fff;border-radius:4px;font-size:.95rem;padding:.4rem .8rem;text-align:center;will-change:transform;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale;background-color:#333}.tippy-tooltip--small{padding:.25rem .5rem;font-size:.8rem}.tippy-tooltip--big{padding:.6rem 1.2rem;font-size:1.2rem}.tippy-tooltip[data-animatefill]{overflow:hidden;background-color:transparent}.tippy-tooltip[data-interactive]{pointer-events:auto}.tippy-tooltip[data-inertia]{transition-timing-function:cubic-bezier(.53,2,.36,.85)}.tippy-tooltip [x-arrow]{position:absolute;width:0;height:0}.tippy-tooltip [x-circle]{position:absolute;will-change:transform;background-color:#333;border-radius:50%;width:130%;width:calc(110% + 2rem);left:50%;top:50%;z-index:-1;overflow:hidden;transition:all ease}.tippy-tooltip [x-circle]:before{content:"";padding-top:90%;float:left}@media (max-width:450px){.tippy-popper{max-width:96%;max-width:calc(100% - 20px)}}
```
maxkfranz: `dist` files probably shouldn't be in the PR.  It's OK to build and test locally, but usually you shouldn't commit dist files (unless you're publishing to npm).

maxkfranz:
@hmtinc I made some initial comments.  Please let me know when you're finished with those, and I'll review a bit more in-depth.  Thanks

hmtinc:
@maxkfranz  I have revised the enhancement and made it so a selector is not required 

#### Sample Usage 
```js
var tip = tippy(refObject, {html: '#template'}); 
```

#### Changes 
* Created a virtual object if a reference object is passed in, so all tippy.js can be used without being modified
* Restored dist files 
* Added checks to ensure certain operations weren't causing errors with the virtual object 
* Removed console.logs used for debugging 
* Resolved issues with getArrayOfElements.js

maxkfranz:
@FezVrasta @atomiks There is a replacement PR that can be cleanly merged in https://github.com/atomiks/tippyjs/pull/112

PR #112, opened 2017-10-10
Added option to override reference object v2

hmtinc:
Note : Previous PR had conflicts due to new commits being made to master. Conflicts have been resolved. 

#### Sample Usage 
```js
var tip = tippy(refObject, {html: '#template'}); 
```

#### Changes 
* Created a virtual object if a reference object is passed in, so all tippy.js can be used without being modified
* Restored dist files 
* Added checks to ensure certain operations weren't causing errors with the virtual object 
* Removed console.logs used for debugging 
* Resolved issues with getArrayOfElements.js

### DIFF
--- a/src/js/tippy.js
+++ b/src/js/tippy.js
@@ -113,7 +113,7 @@ class Tippy {
     const data = find(this.store, data => data.popper === popper)
     const { tooltip, circle, content } = getInnerElements(popper)
 
-    if (!document.body.contains(data.el)) {
+    if (!this.selector.refObj && !document.body.contains(data.el)) {
       this.destroy(popper)
       return
     }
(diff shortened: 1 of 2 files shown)


src/js/tippy.js
```diff
@@ -351,7 +351,36 @@ class Tippy {
...
+        clientWidth: selector.clientWidth,
+        clientHeight: selector.clientHeight,
+        setAttribute: function (key, val) { this.record[key] = val },
+        getAttribute: function (key) { return this.record[key] },
+        removeAttribute: function (key) { return this.record[key] = null },
+        addEventListener: function (key, val) { },
+        removeEventListener: function (key) { },
+        classList : {
+          recordC: {},
+          add : function (key) {this.recordC[key] = true},
```
atomiks: Use ES6 arrow functions indead

src/js/tippy.js
```diff
@@ -351,7 +351,36 @@ class Tippy {
   }
 }
 
+function isElement(element) {
```
atomiks: No need for this function? Simply use `&& !(selector instanceof Element)`

hmtinc:
@atomiks  The requested changes have been made

* Redundant isElement function has been removed
* Selector object now uses ES6 arrow functions

atomiks:
There's been a change to `src/js/core/getArrayOfElements.js`, may you update that file? Then I can merge 😄

atomiks:
Still conflicts?

Also I think yours might break the `NodeList` check since it's an object. I believe a strict object literal check is required...

```javascript
/**
* Returns an array of elements based on the selector input
* @param {String|Element|Element[]|Object|NodeList} selector
* @return {Element[]}
*/
export default function getArrayOfElements(selector) {
  if (selector instanceof Element || Object.prototype.toString.call(selector) === '[object Object]') {
    return [selector]
  }

  if (Array.isArray(selector)) {
    return selector
  }

  if (selector instanceof NodeList) {
    return [].slice.call(selector)
  }

  try {
    return [].slice.call(document.querySelectorAll(selector))
  } catch (_) {
    return []
  }
}
```

Also is the `refObject` in the OP just an empty object literal (since you're overriding it with a virtual one)?

```javascript
var tip = tippy({}, { 
  html: '#template' 
})
```

Just trying to understand this better.

atomiks:
@hmtinc please check pull #116

dgroh:
is there some jsFiddle showing how virtual element works?

PR #113, opened 2017-10-11
Better NodeList Detection

jescalan:
Hi there!

Thanks so much for accepting my previous pull requests! Have been using these features in production and it has gone wonderfully. I noticed when testing in IE that my `NodeList` detection was bugged - unfortunately IE does not have the `constructor.name` property (of course), so if using a `NodeList` input, it would break. This method is much more reliable for detecting whether your input is actually a `NodeList` or not. I tested this in IE and it is working nice and smoothly.

👬

### DIFF
--- a/src/js/core/getArrayOfElements.js
+++ b/src/js/core/getArrayOfElements.js
@@ -12,9 +12,13 @@ export default function getArrayOfElements(selector) {
     return selector
   }
 
-  if (selector.constructor.name === 'NodeList') {
+  if (selector instanceof NodeList) {
     return [].slice.call(selector)
   }
 
-  return [].slice.call(document.querySelectorAll(selector))
+  try {
+    return [].slice.call(document.querySelectorAll(selector))
+  } catch (_) {
+    return []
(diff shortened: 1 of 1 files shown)


atomiks:
How about just `instanceof NodeList`? Seems to work.

jescalan:
Yeah I tested `instanceof` as well and was this close to amending the PR, only left it this way because I'm not sure how good support is for older IE, and I only tested in edge and IE11.

I can run some tests on older versions if you want, or if you don't care about anything below 11 (I don't, as they are well past end of life), I'd be happy to change it over to that.

For the invalid input, want me to have it `console.error` there or throw a more reasonable error message, or would you rather go with the empty array?

atomiks:
Return an empty array (because the `createTooltips` function assumes one) but have this warning before:

`console.warn('Invalid input given to Tippy constructor')`

jescalan:
Sounds good! Just rebased these changes into the original commit to make for a cleaner commit log ✨

PR #116, opened 2017-10-16
refObj implementation

atomiks:
@hmtinc
@maxkfranz

Since there were issues with the pull req I implemented the changes into this branch. Can you verify if this is the correct implementation?

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -128,7 +128,7 @@
             <h4>Dynamic input detection</h4>
             <p>
                 For certain settings to work reliably depending on the type of user input (such as mouse or touch),
-                dynamic input detection is enabled by default. This is only available in v1.0.0+. This can be disabled and enabled at any time via:
+                dynamic input detection is enabled by default. To disable or enable, use:
 
 <pre><code class="language-javascript">tippy.enableDynamicInputDetection()
 tippy.disableDynamicInputDetection()</code></pre>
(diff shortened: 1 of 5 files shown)


hmtinc:
Looks good. Tried it out with the quick demo from the previous pull request and this implementation seems to work.

atomiks:
One last thing, should the `record` property actually be `attributes`?

And `recordC` in `classList` be `className`?

Also to make it work with `title` as well (not require a HTML template), in the refObject you could make it composable

```javascript
const refObject = {
  attributes: {
    title: 'Tooltip text'
  },
  getBoundingClientRect() {
    return {
      top:  5 + window.pageYOffset,
      left: 5 + window.pageXOffset,
      width: 1,
      height: 1
    }
  },
  get clientWidth() {
    return 30
  },
  get clientHeight() {
    return 30
  },
}

// Should work without needing a html setting
tippy(refObject)
```

then `selector` becomes (inside the main `tippy` function):

```javascript
selector = {
  refObj: true,
  attributes: {
    ...(selector.attributes || {})
  },
  etc
}
```

hmtinc:
Yeah, `record` should probably be attributes and `recordC` should be classNames. 
It was initially only named record because it was keeping a record of values,although in hindsight it probably wasn't the best naming scheme .

maxkfranz:
This looks good to me.  

It's great to be able to support the title, as well.

atomiks:
Great! There's a couple more changes I noticed too:

`recordC` also needed `selector.classList.recordC` not just `selector.recordC` inside those `add/remove/contains` functions since that's where it's located in the object.

`removeAttribute` should use `delete` instead of setting it to `null`.

`contains` inside `classList` should return a strict Boolean

The functions should not use implicit returns if possible (setting not getting?)

Here's the final:

```javascript
  if (isObjectLiteral(selector)) {
    selector = {
      refObj: true,
      attributes: selector.attributes || {},
      getBoundingClientRect: selector.getBoundingClientRect,
      clientWidth: selector.clientWidth,
      clientHeight: selector.clientHeight,
      setAttribute: (key, val) => { selector.attributes[key] = val },
      getAttribute: key => selector.attributes[key],
      removeAttribute: key => { delete selector.attributes[key] },
      addEventListener: () => {},
      removeEventListener: () => {},
      classList: {
        classNames: {},
        add: key => { selector.classList.classNames[key] = true },
        remove: key => {
          selector.classList.classNames[key] = false
          return true
        },
        contains: key => !!selector.classList.classNames[key]
      }
    }
  }
```

Can you verify this works with that demo before I finally commit & merge this? Thank you!

hmtinc:
Tested and it works!

PR #119, opened 2017-10-22
Fix followCursor distance

atomiks:
@Whateverinc

Can you verify this fixes the issue?


Whateverinc:
@atomiks Yes, this fixes the issue.

PR #138, opened 2017-12-08
Follow cursor fix

atomiks:
ping @TGotwig

Does this fix the issue?

(`npm run build` this branch and grab the dist file)

### DIFF
--- a/src/js/core/Tippy.js
+++ b/src/js/core/Tippy.js
@@ -294,6 +294,11 @@ export default class Tippy {
         show(event)
       }
       
+      // Set initial positioning near cursor (document mousemove may not fire)
+      if (options.followCursor) {
+        this._followCursorListener(event)
+      }
+      
       // iOS prevents click events from firing
       if (shouldStopEvent && browser.iOS && reference.click) {
         reference.click()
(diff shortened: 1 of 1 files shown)


TGotwig:
No it's like before.

atomiks:
I noticed the 'files changed' doesn't show the diff... Just to verify, is this piece in the build?

```javascript
// Set initial positioning near cursor (document mousemove may not fire)
if (options.followCursor) {
  this._followCursorListener(event)
}
```

TGotwig:
Where these lines of code should be?
Hmm maybe you can reproduce it by using firefox and increase mouse speed 🤔

atomiks:
inside `_getEventListeners()` method in the Tippy class

atomiks:
You're right! Just used Firefox and moved the cursor really slow on your CodePen, and it happened. Let me see if I can reproduce it with this new version...

atomiks:
I believe the latest commit should fix it. It was updating the popper's position when mounting but it shouldn't if the `followCursor` option is true.

TGotwig:
omg it works! 😱😱
Thanks for the fast fix and make this awesome lib even more awesome 💪👌🎉

PR #143, opened 2017-12-13
Add ability to use event delegation

atomiks:
@Prnda1976

Can you check this out and play with it to see if it's all good?

It wasn't as hard as I thought. The `mouseenter focus` gets switched to `mouseover focusin` (for bubbling). The popper instance gets re-used so performance is increased. One slight thing I noticed is that because it's re-used, there can't be two tooltips on the page so it will instantly "hide"/transition out if the cursor went to the next target. Not that big of a deal though...

-----

EDIT: To use dynamic titles:

```html
<div id="parent" title="_">
  <div class="child" data-title="one">text</div>
  <div class="child" data-title="two">text</div>
  <div class="child" data-title="three">text</div>
  <div class="child" data-title="four">text</div>
</div>
```

```js
tippy('#parent', {
  target: '.child',
  onShow() {
    const content = this.querySelector('.tippy-content')
    content.innerHTML = this._reference.getAttribute('data-title')
  }
})
```


rehfeldchris:
I tried your sample, but in the debugger in the onShow() func:

`this._reference` refers to the `#parent` div. Also, `this._reference.dataset` contains originalTitle, and `this._reference.attributes` contains `data-original-title` but not title or data-title. 

In any case, shouldn't we be trying to get the title from the child element?

atomiks:
Hmm I thought I added the code to update the reference upon mount. Lemme see what's wrong..

atomiks:
Demo: https://codepen.io/anon/pen/NXqzYJ

Prnda1976:
@atomiks I have tried to make an example of this on Codepen:

https://codepen.io/Prnda1976/pen/rpVwqd

But for some reason I can't get it to work properly. I get the "tippy is not defined" error even though I'm including:

https://raw.githubusercontent.com/atomiks/tippyjs/target-option/dist/tippy.all.min.js

as external reference. Anyway I didn't really delve deeply into it so perhaps I'm doing something wrong... I will give it another shot during the weekend.

rehfeldchris:
I updated again, this time from 2.0.5 to 2.0.6, verified it via your package.json, but I still have the same issue. Your codepen sample works properly, however.

btw, I'm using angular 4.x, which uses zone.js, in case that might be relevant. I don't have time to make a sample of the app I can share at this moment, but I will come back to it later.

For now:

I did notice that no matter which of the 4 .child I click, that the same Tippy instance is used each time the tip is shown. Eg, Tippy.id = 1, always. And, it's the same Tippy instance, verified by setting a custom property to track it while debugging in the _enter() func.

And, as you prob suspected, I only need to click on the parent to trigger the popover, not necessarily a child.

atomiks:
>But for some reason I can't get it to work properly. I get the "tippy is not defined" error even though I'm including:

Ah, it's because I haven't built the dist files for the branch yet, it's only the src for now. 

>I updated again, this time from 2.0.5 to 2.0.6

Well event delegation isn't in the master branch yet. You need to switch to the "target-option" branch.

rehfeldchris:
I am in that branch.

package.json says
```
{
  "_from": "git+https://github.com/atomiks/tippyjs.git#target-option",
  "_id": "tippy.js@2.0.6",
  "_inBundle": false,
  "_integrity": "sha1-xu5rU7+41A4TYUcr4FhrQFXdaM4=",
  "_location": "/tippy.js",
  "_phantomChildren": {},
  "_requested": {
    "type": "git",
    "raw": "https://github.com/atomiks/tippyjs.git#target-option",
    "rawSpec": "https://github.com/atomiks/tippyjs.git#target-option",
    "saveSpec": "git+https://github.com/atomiks/tippyjs.git#target-option",
    "fetchSpec": "https://github.com/atomiks/tippyjs.git",
    "gitCommittish": "target-option"
  },
  "_requiredBy": [
```

atomiks:
Were you using the dist files? They weren't built yet. Just built them.

The jsfiddle works (manually pasting the built minified file), so not sure why it wouldn't for you?

rehfeldchris:
Yes I was using dist. I did a `npm install https://github.com/atomiks/tippyjs.git#target-option --save` and no src/ files came down, only those in dist/ and a few meta files.

I just updated. It works now :) Thank you so much.

atomiks:
Great! Tell me what bugs you encounter if any. 

I think I need to sort out the accessibility side of things and maybe one or two other options

atomiks:
@Prnda1976 

It appears that GitHub blocks that file from loading in the Codepen, which is why `tippy` isn't defined. Try using the file locally.

rehfeldchris:
I was hoping to be able to set options, like `placement`, when I call the tippy() func to initialize, but also retain the ability to override any option by using the data-tippy-* attributes on the element. Eg, so I can globally use top placement, but for some elements, specify left or right etc...

This seems to work when I don't use the new delegation feature, but not with it. Not sure if that's a bug or a feature request, but I do feel it would be useful to many.

atomiks:
Yes you're right. I forgot about the `data-tippy` attributes for each child target. I will fix that up.

rehfeldchris:
Also, it seems click or focus trigger doesn't work, even on your fiddle.

Thanks!

atomiks:
Yes, they need to be focusable. Adding `tabindex="0"` will fix that

rehfeldchris:
That worked for focus, but click still doesn't work.

atomiks:
Yes I forgot about the `click` trigger 😆 

So I need to:

- Make `click` work
- Make `data-tippy` attributes work
- Make them accessible
- Make `touchHold` option work (don't think it does atm)

rehfeldchris:
It would be nice if the target could be an attribute selector, like `target: '[data-toggle=popover]'`. Currently, it fails due to parsing it as json.

Maybe the fix is as simple as:
`if (typeof val === 'string' && val.trim().charAt(0) === '[' && key !== 'target') {`

atomiks:
Good idea. (had to reset the dist files btw, so they aren't from target-option any more.)

Prnda1976:
@atomiks How is this fiddle example (https://jsfiddle.net/fc1vkvo5/) you show dynamically adding child elements? I mean, they are already there on the page declared inside your html code?

atomiks:
@Prnda1976 you're right it doesn't, but if you modify the code to append new elements to the parent container you'll see it works.

atomiks:
So I just thought of something... because this is proving to be quite difficult (in terms of squashing all bugs and properly setting it up).

Why not use a `MutationObserver` on the parent element to check for subtree changes (i.e. when new elements were appended) and create new tippy instances for the child elements? The downside here is that it's not true event delegation, but it would be way easier to implement properly.

This way you can specify `target` still and still only bind to the parent, but tippy will take care of creating new instances for you.

Thoughts?

rehfeldchris:
I guess the first thing that comes to my mind, being that I'm currently developing w/ angular, which tends to do a lot of dom updates automatically, (not all of which I necessarily want, but it's the price of convenience), it seems like MutationObserver and creating instances would add some additional overhead to the frequent render cycles. 

I'm not personally worried about the performance here, as my needs are pretty simple, but others might.

Have you considered using event delegation of the mouseenter event to identify a tippy element, and then just use that moment to init the instance (if needed)? Instead of creating a single instance, and dynamically updating it with the data like you do now. That might be more compatible with your code, as it's similar to the MutationObserver strategy in that the new feature code is just automatically creating instances, not altering their behavior.

Creating instances on mousenter would be the lazy-load approach, deferring computation until needed. But, MutationObserver would allow pre-initialization, along w/ any benefits that might bring.

atomiks:
`mouseenter` won't work because it doesn't bubble, the child elements don't fire it. `mouseover` is what does that, is that what you mean?

But yes that sounds possible. That way it's still O(1) to init which is great for performance.

rehfeldchris:
Ah, I wasn't aware of that. mouseover then - I'm just speaking high level.

pimmey:
@atomiks hey is this gonna reach master anytime soon? 🙂

atomiks:
@pimmey I got kinda busy recently but might work on it this long weekend and see what I can do. The hard part is integrating it with Popper.js. I might have to completely redo it based on the recent suggestion

atomiks:
See #165

PR #153, opened 2018-01-04
Fix #151

### LINKED ISSUE #151: Does not attach tip when no "title" present, but dynamicTitle is true
The expected behavior for "dynamicTitle" is that it will show a tip once a title is set. However, if the title is not _initially_ set, Tippy won't ever show a tip.

atomiks:
@beether can you verify this fixes the issue?

#151

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,6 @@ module.exports = {
     }
   },
   rules: {
-    indentSwitchCase: true,
     indent: ['error', 2],
     'linebreak-style': ['error', 'unix'],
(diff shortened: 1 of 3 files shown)


beether:
If you can shoot me a URL to the new version I could make a fiddle for you. But right now am too busy to do a git clone and all of that... Thanks for the quick response to this though! I've been loving the library.

atomiks:
Yeah here https://jsfiddle.net/syp1pgz0/

atomiks:
@beether?

From what I can see in the fiddle and my own testing, it appears to work properly

PR #165, opened 2018-01-26
Add ability to use event delegation

atomiks:
Defers the creation of individual Tippy instances for each target now. Much better method of doing it and fixes every issue automatically.

```html
<div id="parent" title="Shared">
  <div class="child">text</div>
  <div class="child">text</div>
  <div class="child" title="Own">text</div>
</div>
```

```js
tippy('#parent', { target: '.child' })
```

The parent wrapper doesn't need a `title` attribute, and if it does it gets shared across each child. Can easily be overridden by specifying a `title` attribute on the child element.


nchaud:
Just a usage note (and perhaps a tweak required?) - this only works if the title or data-title attribute is present. If you're using the html property to show content based on a template, this won't work. 

Workaround: Add a dummy title attribute.

PR #173, opened 2018-01-30
fix #170

### LINKED ISSUE #170: disable focus on interactive popover
Please add option to disable it
`if (options.interactive && elementIsInViewport(reference)) {
								popper.focus();
							}`

atomiks:
@fenyagg

This should fix the issue, can you confirm?


fenyagg:
everything is fine!

atomiks:
Awesome, I made it a function and removed the `elementIsInViewport` util as that was an inferior method to what I just made. Thanks for making the issue 😄

PR #180, opened 2018-02-10
fix #179

### LINKED ISSUE #179: Each tooltip starts offset to bottom left and first trigger transitions it into position
I'm seeing the following buggy behavior on all tooltips, both in v1 and now v2:

It seems every tooltip instance is somehow first transitioned from its initial position towards the target center, but not anymore once it has been loaded at least once.

atomiks:
@mystrdat

Before making `updateDuration` 0 by default, does this fix the issue? There's two updates on show due to a problem with arrow positioning which may be causing the transition on show issue, although I doubt it.

### DIFF
--- a/src/js/core/Tippy.js
+++ b/src/js/core/Tippy.js
@@ -4,6 +4,7 @@ import Popper from 'popper.js'
 
 import tippy from '../tippy'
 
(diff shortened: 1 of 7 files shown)


mystrdat:
Sorry for not confirming earlier, looks fixed!

PR #192, opened 2018-02-26
Remove the word "simply" from documentation

adactio:
The word "simply" (like the word "just") is probably intended to convey that the instructions are straightforward, but it can have the unintended side-effect of making the reader feel overwhelmed:

http://bradfrost.com/blog/post/just/


atomiks:
Interesting, didn't think of it like that. Thanks for the contribution ✌️

PR #194, opened 2018-02-27
Hiding poppers no longer throws errors if tippy is undefined

divmgl:


atomiks:
When would the popper not have have `_tippy` attached?

divmgl:
@atomiks It's happening right now in a production app. I was able to remedy it by first destroying the popper then destroying the tippy instance but it's definitely possible for some weird reason. In a React app, if the popper is open and you perform a click event that causes the parent component to unmount, an exception will be thrown.

atomiks:
Pretty weird but it's not like this affects anything so it's fine 😃

Issue #261, opened 2018-07-12
removeOnDestroy option issue


worldofsites:
When we are testing tippyjs on **IE11**, we have found that after call .destroy() or .destroyAll() div's (.tippy-popper) doesn't removed from DOM. Even if we create tippy with 'removeOnDestroy: true' option.
![image](https://user-images.githubusercontent.com/23352829/42612146-d44f1110-85c3-11e8-8a26-3c67ced6e352.png)
Issue occurs because in function _createPopperInstance() option removeOnDestroy doesn't transfered to popper instance options.
**Possible fix:**
Adding 'removeOnDestroy: options.removeOnDestroy' to _extends in _createPopperInstance() function is solved that issue.
![image](https://user-images.githubusercontent.com/23352829/42612141-cae30fb4-85c3-11e8-8140-71290248d718.png)

atomiks:
Was fixed in v2.5.4 with the `transitionend` change.

PR #262, opened 2018-07-12
 fix for removeOnDestroy option

worldofsites:
FOR https://github.com/atomiks/tippyjs/issues/261


atomiks:
See #264, this would be the issue with IE11. Surprised this has only been reported now

Issue #263, opened 2018-07-13
onHidden and onShown not triggered in IE11


DaJoTo:
Only wait, onHide and onShow callbacks are being called within IE 11. On further investigation found the issue to be checking for `ontransitionend in window`.

PR #264, opened 2018-07-13
Modify how transitionend listener is determined

DaJoTo:
FOR #263

This PR provides a function to safely determine the correct transitionend listener to use for the active browser. This fixes an issue with IE11 where onHidden and onShown are not being called.

### DIFF
--- a/src/js/core/Tippy.js
+++ b/src/js/core/Tippy.js
@@ -716,7 +716,9 @@ export function _onTransitionEnd(duration, callback) {
   const toggleListeners = (action, listener) => {
     if (!listener) return
     tooltip[action + 'EventListener'](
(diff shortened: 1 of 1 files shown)


atomiks:
Thank you, but can you only edit the `src` not the `dist`.

Also, `document.body` is assumed throughout the library so use that instead of the mock element

atomiks:
Can you verify that `'ontransitionend' in document.body` this works in IE11? we don't need to check for non-webkit prefixes.

atomiks:
Can you verify this works?

```js
'transition' in document.body ? 'transitionend' : 'webkitTransitionEnd'
```

We don't need to check for non-webkit browsers

DaJoTo:
I'll just correct and check it now, but don't see why it wouldn't.

DaJoTo:
yep, all good, have updated

PR #289, opened 2018-09-06
add useCapture static method for global click listener

atomiks:
@gargrave. How's this?

```js
// Call this before invoking `tippy()`
tippy.useCapture()
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 .DS_Store
 /node_modules
-dist/tests
+dist
+coverage
 .cache/
+package-lock.json
+.devserver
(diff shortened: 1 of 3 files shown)


gargrave:
Nice, that should work! I'll test it out first thing in the morning and back to you. Thanks a bunch! 👍

gargrave:
This works perfectly, and is _exactly_ what we needed. As far as my use case goes, this is 100%--nothing more needed.

Thank you so much for your responsiveness and flexibility here--this little tiny fix solves a pretty big issue for us, and I can't express enough gratitude for how quick and easy you made this for us. 🥇

atomiks:
Great! The test only failed because of jest v22, v23 has been updated on master which fixes the issue.

PR #307, opened 2018-10-01
Use textContent instead of innerHTML to inject CSS

marco-c:


atomiks:
Good catch, didn't think about it in this case. Does it work fine on IE11 etc?

marco-c:
I haven't tested it on IE11 yet, I can try in a bit.

atomiks:
I've tested IE11 using Browserstack + Chrome, FF, Safari, iOS, etc. and it's good.

marco-c:
Thanks for testing it for me ;)

atomiks:
Apologies for the wait: released in `3.0.4`

PR #315, opened 2018-10-09
Fix followCursor jumpiness with fast transition duration

atomiks:
@kfir124

From my testing this seems to fix the issue by replacing `scheduleUpdate` with `update` without performance impact, can you verify?

### DIFF
--- a/src/js/createTippy.js
+++ b/src/js/createTippy.js
@@ -156,7 +156,7 @@ export default function createTippy(reference, collectionProps) {
       clientHeight: 0
     }
 
-    tip.popperInstance.scheduleUpdate()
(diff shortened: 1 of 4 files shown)


kfir124:
It seems to work! 👍 

Couldn't really check the performance hit (if any) but I couldn't reproduce the jump bug

atomiks:
I believe it could've still potentially happened, just even rarer than before.

I just refactored to the code to 100% prevent the race condition from occuring. Mind checking the latest commit?

kfir124:
Checked and it seems to work great! didn't encounter the jumpy bug

PR #318, opened 2018-10-10
fix setContent() resetting followCursor position

atomiks:
cc @NilsEnevoldsen


atomiks:
Note that this means if you change one of the whitelist options in `set()` the position still gets reset. But those aren't common things to change in combination with `followCursor`.

NilsEnevoldsen:
I realize my emoji reaction may not have sufficiently conveyed my appreciation. Thank you! This seems like a good solution. Could it make it into a 3.0.6 release? 😄

atomiks:
Yeah in a sec, just solved the caveat above. I'm driving somewhere but I'll get it out in an hour or so

PR #324, opened 2018-10-16
Update incorrect option table entries in documentation

KubaJastrz:


atomiks:
Thank you for spotting it!

Issue #186, opened 2018-02-18
Need a typescript definitely typed file available in npm package


vairamsvsjdo:
@artyomtrityak 
Is there a typescript definitely typed file available for this library?

renehamburger:
I don't have time for a PR right now, but here is an up-to-date d.ts file for tippy.js I created yesterday. It would be create if it could be included directly in this library (rather than in the [DefinitelyTyped repository](https://github.com/DefinitelyTyped/DefinitelyTyped)). It may be possible to take the popper definitions directly [from here from the popper package](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/index.d.ts). I adjusted them slightly by adding a namespace.

### tippy.d.ts
```ts
// Type definitions for tippy.js v2.2.3
// Project: https://atomiks.github.io/tippyjs
// Definitions by: renehamburger <https://github.com/renehamburger>

///<reference path="./popper.d.ts" />

declare namespace Tippy {

  type Placement = 'top' | 'bottom' | 'left' | 'right';

  type Trigger = 'mouseenter' | 'focus' | 'click';

  /** @see https://atomiks.github.io/tippyjs/#all-options */
  interface Options {
    /**
     * Specifies which direction to place the tooltip in relation to the reference
     * element. Add the suffix `-start` or `-end` to shift the placement, e.g. `'top-end'`.
     * Default: 'top'
     */
    placement: Placement;
    /**
     * If `false`, the tooltip won't flip when scrolling. This was unconfigurable and
     * always `true` before v2.2.
     * Default: true
     */
    livePlacement: boolean;
    /**
     * Specifies which type of events will trigger a tooltip to show. Separate each
     * by a space. `mouseenter` is for hovering and touch on mobile, and `focus` is for
     * keyboard navigation. Use `manual` if you want to show/hide the tooltip manually
     * (see https://atomiks.github.io/tippyjs/#methods). You can also use any other event listener,
     * but it won't have the opposite "hide" event.
     * Default: 'mouseenter focus'
     */
    trigger: Trigger | 'manual' | string;
    /**
     * Whenever the `title` attribute on the reference element changes, the tooltip will
     * automatically be updated.
     * Default: false
     */
    dynamicTitle: boolean;
    /**
     * Makes a tooltip interactive, i.e. will not close when the user hovers over or
     * clicks on the tooltip. This lets you create a popover (similar to Bootstrap)
     * when used in conjunction with a `click` trigger.
     * Default: false
     */
[...]

PR #329, opened 2018-10-19
Introduce types file for TypeScript module support

KubaJastrz:
Basically, I've created `index.d.ts` in the main directory with a tweaked version of @smajl typings found in #186 discussion.

If you want to use the standalone file instead of the default one you can declare a module in `*.d.ts` file inside your project like this:
```ts
declare module 'tippy.js/dist/tippy.standalone' {
  import { Tippy } from 'tippy.js';
  const tippy: Tippy;
  export default tippy;
}
```

If you have any suggestions, please let me know :slightly_smiling_face: 
Before merging I could probably add comments to each definition.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,112 @@
+import Popper from 'popper.js'
+
+export type BasicPlacement = 'top' | 'bottom' | 'left' | 'right'
+
+export type Placement =
+  | BasicPlacement
+  | 'top-start'
+  | 'top-end'
+  | 'bottom-start'
+  | 'bottom-end'
+  | 'left-start'
+  | 'left-end'
+  | 'right-start'
+  | 'right-end'
+
+export type Content = string | Element
+
+export type Selector = string | Element | NodeList | Popper.ReferenceObject
+
+export interface Options {
+  a11y?: boolean
+  allowHTML?: boolean
+  animateFill?: boolean
+  animation?: 'fade' | 'scale' | 'shift-toward' | 'perspective' | 'shift-away'
+  appendTo?: Element | ((ref: Element) => Element)
+  arrow?: boolean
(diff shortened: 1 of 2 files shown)


atomiks:
Thanks for this!

> If you want to use the standalone file instead of the default one you can declare a module in *.d.ts file inside your project like this:

The `"module"` field points to the standalone file in package.json though?

KubaJastrz:
Yes. `"module"` field is not officially included in [package.json docs](https://docs.npmjs.com/files/package.json) yet, but is supported by modern packagers, so I just left it there if someone needs it (like when [importing different file](https://atomiks.github.io/tippyjs/#getting-started-files)). It should work out of the box for most use cases though.

atomiks:
A couple changes:

- `Selector` should be `Targets`, as a CSS selector is not the only type
- `Options` - should it be `Props`?

Regarding the 2nd one, my thinking is this: `props` is the object of properties used to configure the tippy. They are called `options` if they can be optionally supplied to the tippy function.

So it would be like this (example):

```js
one(selector: Selector, options?: Props): Instance
```

KubaJastrz:
Can `targets` argument be an array of some sort? I found this bit of code and I'm wondering if `Array.isArray(value)` should be validating only `Element` array instead of any array. https://github.com/atomiks/tippyjs/blob/c8497548becb464d7113581f0488c5421e01950a/src/js/utils.js#L104-L120

atomiks:
I think it's good now. 

Should this be released in v3.1? I wanted to add another theme to `dist/themes`, more properties to `state`: `isShown` and `isMounted`, a new lifecycle function `onMount`, and some new animations in `dist/animations`.

KubaJastrz:
I guess it's fine. As soon as it is released I'll probably work on typings for the react component.

PR #347, opened 2018-11-07
Add v-tippy plugin to wrappers list

andreasvirkus:
Wondering if you'd also be willing to list community plugins for Tippy?
I've listed one from egoist (Vue core contributor and a Vue community "giant").

Best,
Andreas

### DIFF
--- a/website/js/sections/GettingStarted.js
+++ b/website/js/sections/GettingStarted.js
@@ -63,6 +63,11 @@ export default () => (
           React
         </ExternalLink>
       </li>
+      <li>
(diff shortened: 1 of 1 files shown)


atomiks:
I'm only going to put official wrappers in that list (because that one is v2 instead of v3, and they can go unmaintained easily). All wrappers are published under the `@tippy.js` organisation, I'd like an official `@tippy.js/vue` package, along with the other view libraries too.

Do you know how you'd make one similar to the React wrapper for Vue? I'm not familiar with it.

andreasvirkus:
I can look into it tomorrow night :)

Issue #355, opened 2018-11-16
Include SCSS files in release


davefmoore:
It would be awesome to have access to the scss files, ability to overwrite variables (ie colors) to match theme.

nikolowry:
Somewhat off-topic, but injecting styles into the head as the first child produces the following W3C validation error: `Error: A charset attribute on a meta element found after the first 1024 bytes`.

It be great if we had the ability to disable rogue injections into the DOM

atomiks:
@nikolowry this might work instead

```js
const firstStyleOrLinkElement = head.querySelector('style,link')

if (firstStyleOrLinkElement) {
  head.insertBefore(style, firstStyleOrLinkElement)
} else {
  head.appendChild(style)
}
```

atomiks:
Don't think this is happening 🙈

PR #356, opened 2018-11-17
 Include scss files in distribution

sqal:
Resolves https://github.com/atomiks/tippyjs/issues/355. I've also added `$tippy-` prefix to all variables so that they do not interfere with other user's variables.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,7 +2,7 @@ docs
 node_modules
 scripts
 tests
-src
+src/js
 build
 .cache
 .github
--- a/src/scss/_vars.scss
+++ b/src/scss/_vars.scss
@@ -1,4 +1,4 @@
-$tippy-placements: 'top', 'bottom', 'left', 'right';
-$tippy-origins: bottom, top, right, left;
-$tippy-backdrop-origins: 0% 25%, 0% -50%, 50% 0%, -50% 0%;
-$tippy-backdrop-border-radii: 40% 40% 0 0, 0 0 30% 30%, 50% 0 0 50%, 0 50% 50% 0;
+$tippy-placements: 'top', 'bottom', 'left', 'right' !default;
+$tippy-origins: bottom, top, right, left !default;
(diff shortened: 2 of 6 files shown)


atomiks:
Can I ask how you'd use these files in an example? Trying to contextualize it a bit

sqal:
@atomiks You would import them this way:

// variable overrides goes before @import

@import "tippyjs/src/scss/tippy";
@import "tippyjs/src/scss/tippy/themes/translucent";

Also, what i forgot to add in this PR is the ability to overwrite theme's variables. I'd like to hear, how would you prefer to have it done? There are ways to do this.

1. Using multiple variables with theme's name:

```scss
$tippy-theme-google-color: #505355 !default;
$tippy-theme-google-color: white !default;
```

2. Or imho, better approach is using maps

```scss
$tippy-theme-google: () !default;
$tippy-theme-google: map-merge(
  (
    color: #505355,
    background-color: white
  ),
  $tippy-theme-google
);
```

Either way, user can easily change any variable's value.

KubaJastrz:
Is it a common practice to include source code in the npm package? This looks fine, but I have mixed feelings about it.

Either way, if we want to implement this, it would be a good idea to include above instructions in the docs somehow.

ZhangYiJiang:
@KubaJastrz It is quite common, especially for frameworks - 

- Bootstrap: https://getbootstrap.com/docs/4.0/getting-started/theming/ 
- Material Components Web - https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md#step-2-include-css-for-a-component 

The benefits of this is that it enables better, smaller builds since as end users we can use our own autoprefixer/PostCSS configs, allowing us to create smaller builds, and makes it easier to theme the styles as mentioned above

atomiks:
I think we need to also do this for animations and themes then.

Also I'd like to figure out a good CSS-in-JS solution for V4 so you don't need to import a CSS file for React

atomiks:
Happy to accept a new PR that does it for themes etc. too!

PR #362, opened 2018-11-23
Reintroduce maxWidth option

### LINKED ISSUE #361: Is maxWidth still valid?
I need to set a width on the tooltips, failing that a maximum width.

When trying to use `maxWidth` I get the following error:

`Error: [tippy]: maxWidth is not a valid option`

Has it been removed? If so, is there an alternative? You can't style reliably with simple css as it then throws positioning out.

atomiks:
Resolves #361 

We remove the media query and place this property in `.tippy-popper`

```css
.tippy-popper {
  max-width: calc(100% - 10px);
}
```

It prevents the tooltip from ever overflowing if it's too large due to its content, even if style.maxWidth is specified.

Also unlike in V2 you can just use a number without a unit and "px" gets added.

Thoughts? /cc @darrensw777

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -45,6 +45,7 @@ export interface Props {
   interactiveDebounce?: number
   lazy?: boolean
   livePlacement?: boolean
+  maxWidth?: number | string
   multiple?: boolean
   offset?: number | string
   onHidden?(instance: Instance): void
--- a/src/js/defaults.js
+++ b/src/js/defaults.js
@@ -21,6 +21,7 @@ export let Defaults = {
   interactiveDebounce: 0,
   lazy: true,
   livePlacement: true,
+  maxWidth: '',
   multiple: false,
   offset: 0,
   onHidden() {},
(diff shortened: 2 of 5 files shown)


darrensw777:
That sounds great, back where it was - you're right in your previous comment on #361 though, a theme could also achieve this I just didn't want to go down that route ideally. Thanks for the quick attention, much appreciated!

KubaJastrz:
Is it even needed to have it when popper's [`preventOverflow.padding`](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.padding) is available?

atomiks:
@KubaJastrz I believe that just specifies how much padding between the boundary and the popper there is, but if the popper is say 1000px wide and the screen is 500px wide, then there's nothing done to minimize the width, as that's defined by CSS

KubaJastrz:
Oh, okay. I wasn't sure if popper takes care of _fitting_ an object in the boundary. Makes sense now.

atomiks:
> EDIT: I mean as a replacement for old media query. It can be useful as a standalone option.

I agree it would be useful as a base-level option (not needing to use `popperOptions`). I can make a different PR for it

Issue #272, opened 2018-08-08
How to position in the mouse click?


dgroh:
I finally managed to position using pure `popper` a div inside a SVG. I wish to do the same with `tippy` 

I'm trying the following approach which **does not work**:

```
  select(e: MouseEvent) {
    const ref = {
      getBoundingClientRect() {
        return {
          width: e.layerY,
          height: e.layerX,
          top: e.layerY,
          left: e.layerX,
          right: 0,
          bottom: 0
        }
      },
      clientHeight: 100,
      clientWidth: 100
    }

    tippy('.line', {
      target: 'polyline',
      trigger: 'click',
      arrow: true
    }, ref);
  }
```

**This worked** with popper:

```
  select(e: MouseEvent) {
    const ref = {
      getBoundingClientRect: () => ({
        top: e.layerY,
        right: e.layerX,
        bottom: e.layerY,
        left: e.layerX,
        width: 0,
        height: 0,
      }),
      clientWidth: 0,
      clientHeight: 0,
    }
    
    const pop = document.getElementById('template'); // the popup element

    this.popper = new Popper(ref, pop);    
  } 
```

What am I missing?

davemoore24:
The reference is virtual so `.line` needs to be replaced with `ref`

`   const ref = {
        getBoundingClientRect: function getBoundingClientRect() {
          return {
            width: 0,
            height: 0,
            top: clientY,
            left: clientX,
            right: clientX,
            bottom: clientY
          };
        },
        clientWidth: 0,
        clientHeight: 0
      };

      tippy(ref, {
        target: 'polyline',
        trigger: 'click',
        arrow: true
      });`

atomiks:
Yeah, you need to use the object as the actual reference itself.

But I see that you want to use the line as the reference for tippy to handle listeners for, but want a custom position where they clicked... that's a really valid use case that hasn't been added yet 😢 .

dgroh:
That's is correct! I did it without the use of a library. I used this: http://www.cssarrowplease.com/ , then simply applied **absolute** positioning with **top** and **left**. pretty simple, but covers my use case for now.

viveleroi:
I too need to be able to position the tooltip relative to the cursor, but *not* follow it. Is that what this issue is trying to address?

I would try the virtual ref approach but I need this to work with `dynamicTitle` too.

atomiks:
@viveleroi Yes, it seems so.

[...]

PR #374, opened 2018-12-03
Add followCursor: 'initial' option

### LINKED ISSUE #373: Feature Request: followCursor once
I would like the pop-up to appear near the cursor, but not follow it. In my situation, I have a pop-up associated with a row in a table that is the width of the screen. I'd like the pop-up to be near the mouse, but not jump around once it appears.

Maybe a new option 

followCursor: "initial"

or something like that. Thanks for a great little package!

atomiks:
Resolves #373 

Check it out @dkulp2?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -37,7 +37,7 @@ export interface Props {
   distance?: number
   flip?: boolean
   flipBehavior?: 'flip' | Placement[]
-  followCursor?: boolean | 'vertical' | 'horizontal'
+  followCursor?: boolean | 'vertical' | 'horizontal' | 'initial'
   hideOnClick?: boolean | 'toggle'
   inertia?: boolean
   interactive?: boolean
--- a/src/js/createTippy.js
+++ b/src/js/createTippy.js
@@ -243,6 +243,10 @@ export default function createTippy(reference, collectionProps) {
     }
 
     tip.popperInstance.scheduleUpdate()
+
+    if (followCursor === 'initial' && tip.state.isVisible) {
+      removeFollowCursorListener()
+    }
   }
 
   /**


KubaJastrz:
Please, add that option to types file.

dkulp2:
As long as the listener is sure to be attached with that value to begin with, looks great to me!

atomiks:
> As long as the listener is sure to be attached with that value to begin with, looks great to me!

Yeah it just checks for a truthy value to determine if it should be attached. It think this is the same as the `nearCursor` option that was proposed a few months ago, for some reason I didn't think it would be so simple

atomiks:
This allows you to implement "native" (at least on macOS) tooltip behavior. It will show up at a `bottom-start` placement at the initial cursor position. Cool feature I think 😌

clbn:
@atomiks It seems this also closes #272?

atomiks:
Yeah it seems so, lol

clbn:
@atomiks
By the way, is there a way to combine `followCursor: 'initial'` and `placement: 'bottom'`? I need this feature now but couldn't find the way. So the tooltip would show up at mouse's horizontal position, and using the trigger's `bottom` for vertical coordinate. For "wide table example", this would be quite useful as it would not cover the row itself and look more "orderly".

atomiks:
I think I know what you mean, you don't want it to vertically overlap but still position itself at the curosr on the horizontal axis.

I was working on that for inline elements specifically, but inline positioning might need to be integrated into Popper.js itself 

> Popper.js does not support smart inline element positioning. You can view [one technique here](https://codepen.io/atomiks/pen/GeBoyW?editors=0010)

(there used to be a bug with flipping in some cases, but I can't replicate it anymore)

You can modify it so it uses the event coords for left/right: https://codepen.io/atomiks/pen/GRKGwQa?editors=0110

Issue #392, opened 2019-01-03
Support aria-labelledby


jschuler:
There are cases where a tooltip is serving as the label for something, like an icon that isn't obvious. In this case, the better attribute to use would be aria-labelledby, which is announced regardless of navigation method (tab key, arrow key, ...). Would be great if I had the choice to choose between aria-describedby and aria-labelledby

atomiks:
`3.4.0`

Issue #391, opened 2019-01-03
Interactive tooltip should not be auto-focused


jschuler:
I want to have a tooltip that remains open when i hover away from the reference element and hover over the tooltip. I can do this by setting interactive option to true.
However, I noticed that when this option is set to true, the tooltip will receive focus as soon as it opens.
This causes an a11y issue with screen readers, as the reading of the reference element text is interrupted by the tooltip text.
I'd like to keep focus on the reference element.

Example: https://codepen.io/jschuler/pen/maXqZQ

jgiardino:
Adding to @jschuler's description, the behavior he describes is based on [WCAG 2.1 item 1.4.13](https://www.w3.org/WAI/WCAG21/quickref/#content-on-hover-or-focus):
> Hoverable: If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

Also, the screen reader issue he describes was noticed in Voiceover on Mac.

atomiks:
Hm I assumed that the reference element would finish being read before swapping to the tooltip. I guess not.

How should we handle focus inside the tooltip but still allow the reference element to be read out (with `focus` trigger). It seems `click` is a lot more accessible.

atomiks:
`.3.4.0`

atomiks:
`3.4.0`

PR #393, opened 2019-01-04
`autoFocus` and `aria` options

atomiks:
cc @jschuler, @jgiardino

### `autoFocus: true`

Determines if interactive poppers should be auto focused programmatically

### `aria: 'describedby'`

Either 'describedby' or labelledby'

Does this look appropriate?


KubaJastrz:
resolves #391, resolves #392

PR #394, opened 2019-01-04
Add 'boundary' option

atomiks:
Flat-level alias for `popperOptions.modifiers.preventOverflow.boundariesElement` since it's common

### DIFF
--- a/src/js/createTippy.js
+++ b/src/js/createTippy.js
@@ -19,7 +19,14 @@ import { canReceiveFocus } from './reference'
 import { validateOptions, evaluateProps } from './props'
 import computeArrowTransform from './deprecated_computeArrowTransform'
 import { closest, closestCallback, arrayFrom } from './ponyfills'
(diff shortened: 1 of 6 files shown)


src/js/createTippy.js
```diff
@@ -476,20 +476,23 @@ export default function createTippy(reference, collectionProps) {
...
+    return new Popper(tip.reference, tip.popper, {
       placement: tip.props.placement,
       ...(popperOptions || {}),
       modifiers: {
         ...(popperOptions ? popperOptions.modifiers : {}),
+        preventOverflow: {
+          boundariesElement: tip.props.boundary,
+          ...(popperOptions && popperOptions.modifiers
+            ? popperOptions.modifiers.preventOverflow
+            : {}),
```
KubaJastrz: Maybe you can create a helper for extracting `popperOptions`? I can see this patterns a few times here.

atomiks: Good point

PR #395, opened 2019-01-04
Typings improvements

KubaJastrz:
Previous value allows only one trigger to be available, you couldn't create string combinations like: `mouseenter focus` etc.

This is a quick fix, the best option is to include props descriptions copied from the docs.


atomiks:
Also you can return `false` from `onShow` and `onHide`, can that be fixed?

PR #396, opened 2019-01-05
New website using Gatsby

atomiks:
Using Gatsby + MDX for content, improved DX and UX I think 😎 

Content is split across several pages with PREV / NEXT buttons at the bottom and a sidebar for navigation. I'll probably continue spicing it up over time but it's decent enough for now, and an improvement over the original in terms of content consumption.


KubaJastrz:
Great work!

PR #413, opened 2019-01-23
V4 (WIP)

atomiks:
### Core

- Rename `performance` option to `ignoreAttributes`
- Remove `arrowTransform` option
- Remove `tippy.useCapture()` & `tippy.disableAnimations()`
- Remove `tippy.one()` and collections, return the instance directly if Element or VirtualElement, otherwise array of the instances
- `tippy.hideAllPoppers()` renamed to `tippy.hideAll()` and ignores `hideOnClick` value for static method consumers
- CSS simplified, opinionated CSS like font-smoothing removed
- `tippy.group()` method
- In event delegation mode, `content` and `appendTo` as functions receive the target element rather than the delegate reference
- `updateDuration: 0` by default so flips don't have the weird transition, and specifying any number uses true smooth flips (with no problem on IE11 now)
- `MutationObserver` removed, `.set()` calls `scheduleUpdate()` now, and if changing a HTML element as the content, the user should manually call `.scheduleUpdate()` or use a `ResizeObserver` polyfill for complete update accuracy
- Flipping has changed. `livePlacement` was renamed to `flipScroll` and now position _does_ still update on scroll, but it doesn't flip while scrolling. Flipping while scrolling is now disabled by default since it's better UX, I think.
- Remove all focus handling: leave it up to the developer. Recommended option requires no focus handling: append it directly after the trigger so elements can be tabbed to. There may be some stacking issues where it needs to be appended to the body, but they can programmatically .focus() it if need be and make it act like a modal. 

### Package

- Babel 7
- `popper.js` is external in all files and formats
- 4 formats:
-- `index.js` = tippy
-- `index.min.js` = tippy minified
-- `index.all.js` = tippy + css
-- `index.all.min.js` = tippy + css minified
- Files are no longer nested under `dist` and exist in the root directory, aka

```
tippy.js/
├── umd/
│   ├── index.js
│   ├── ...
├── esm/
│   ├── index.js
│   ├── ...
├── themes/
│   ├── light.css
│   ├── ...
├── index.css
├── index.d.ts
├── LICENSE
├── README.md
├── package.json
```

I think we should default to the injectCSS method for modules as well, no need to import CSS manually - thoughts?:

```json
{
  "main": "./umd/index.all.min.js",
  "module": "./esm/index.all.js"
}
```

Use minified code for `"main"` field (since sourcemaps anyway?), so CDN include looks like this:

```html
<script src="https://unpkg.com/popper.js@1/dist/umd/popper.min.js"></script>
<script src="https://unpkg.com/tippy.js@4"></script>
```

Examples:

```js
// A: Uses esm format + injects CSS
import tippy from 'tippy.js'

// B: Uses cjs format + import css
const tippy = require('tippy.js/umd/index.js')
require('tippy.js/index.css')
```

Minzipped bundle sizes look like:

JS: `5.93 kB`
CSS: `1.56 kB`

/cc @KubaJastrz


atomiks:
Any suggestions for a JSS theming system for React et al? I would like V4 to be themeable in JS in some way to be used in CSS modules and `styled-components` without using global styles. Or is it fine as it is? https://github.com/atomiks/tippy.js-react/issues/14

e.g. inline styles - the main problem is styling for different `x-placement`s

```js
tippy(ref, {
  theme: {
    arrow: { borderColor: 'white' }, // how would this work for diff x-placements? 
    tooltip: { background: 'white' }
  }
})
```

KubaJastrz:
Maybe make it a function and pass some props to it?
```js
tippy(ref, {
  theme: ({ position }) => ({
    arrow: { borderColor: position === 'left' ? 'white' : 'black' }, 
    tooltip: { background: 'white' }
  })
})
```
Or it can be both :thinking:

atomiks:
Yeah that seems good actually.

So it should be the base placement, ignoring shifting?

```js
const capitalize = str => placement[0].toUpperCase() + placement.slice(1)

tippy(ref, {
  theme: ({ placement }) => ({
    arrow: { [`border${capitalize(placement)}Color`]: 'white' },
  })
})
```

atomiks:
`tippy.hideAllPoppers` is renamed to `tippy.hideAll`.

Also it can now take an options object:

```js
tippy.hideAll({ duration: 0 })
tippy.hideAll({ exclude: instance })
```

@KubaJastrz how is `index.d.ts` looking?

index.d.ts
```diff
@@ -102,22 +101,18 @@ export interface Instance {
...
 
 export interface Tippy {
-  (targets: Targets, options?: Props): Collection
+  (targets: Targets, options?: Props): Instance | Instance[]
   readonly defaults: Props
   readonly version: string
-  disableAnimations(): void
-  hideAllPoppers(): void
-  one(targets: Targets, options?: Props): Instance
+  hideAll(options: HideAllOptions): void
```
KubaJastrz:
```suggestion
  hideAll(options?: HideAllOptions): void
```
Otherwise, it will require an object to be passed.

KubaJastrz:
And I think you missed `performance` to `ignoreAttributes` change.

atomiks:
How should we include the [`group`](https://codepen.io/anon/pen/mQKjrv?editors=1010) function? Like this?

### UMD

```html
<script src="https://unpkg.com/popper.js"></script>
<script src="https://unpkg.com/tippy.js"></script>
<script src="https://unpkg.com/tippy.js/plugins/group"></script>
<!-- 
All plugins are at the index.js file:
<script src="https://unpkg.com/tippy.js/plugins"></script> 
-->

<script>
  const instances = tippy('button')
  tippy.group(instances, { delay: 1000 })
</script>
```

### ES

```js
import tippy from 'tippy.js'
import { tippyGroup } from 'tippy.js/plugins'

const instances = tippy('button')
tippyGroup(instances, { delay: 1000 })
```

KubaJastrz:
Interesting idea with plugins, I'm not sure about creating separate requests for each one though when importing using `<script>` tags. With [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) it wouldn't be an issue, but importing few small files instead of one bigger may lead to opposite effect.

I suggest one bundle that contains bare minimum and is preferred for most use cases and one bigger with all plugins. I guess it doesn't hurt to allow users to have a choice and build one of each:
- have one small package and optional plugins as suggested above
```html
<script src="https://unpkg.com/tippy.js"></script>
<script src="https://unpkg.com/tippy.js/plugins/group.js"></script>
<script src="https://unpkg.com/tippy.js/plugins/some-other.js"></script>
```
- and one big package that includes everything (except css)
```html
<script src="https://unpkg.com/tippy.js/with-plugins.js"></script>
```

atomiks:
A HTTP request is probably more expensive than extra unused bytes indeed, but `themes` currently operate that way and the package _is_ more optimized for a modern build chain with bundlers than via a CDN

I guess a "plugin" in a traditional sense will attach itself to the main namespace, so both esm/umd versions should do that rather than being a separate module?

group.js

```js
tippy.group = function group() { /* ... */ }
```

example App.js

```js
import tippy from 'tippy.js'
import 'tippy.js/plugins/group' // must be imported after main import

tippy.group(...)
```

This harmonizes CDN + bundler usage, but most modern packages use separate exports instead. For a CDN, that means creating a new namespace, `tippyGroup` instead of attaching it to the main one.

KubaJastrz:
I like attaching to the `tippy` object. Actually, there's no need for another namespace at all. What I would do is to have already preloaded plugins for the CDN that would be created like this:
```js
import tippy from 'tippy.js'
import 'tippy.js/plugins/group' // appends .group method to main object

export default tippy
```

atomiks:
Couldn't really get that functionality working sanely, it seems like a legacy thing to do... given that it's only 0.2 kB minzipped, I'll just include it by default 😅

atomiks:
I'll release an alpha soon so people can test it out

src/js/utils.js
```diff
@@ -118,3 +89,39 @@ export function getModifier(obj, key) {
...
+  return isPlainObject(value) || value instanceof Element
+}
+
+/**
+ * Tricking bundlers, linters, and minifiers
+ * @return {String}
+ */
+export function innerHTML() {
+  return 'innerHTML'
+}
```
KubaJastrz: I'm still not sure why this is needed

atomiks: It's because Firefox extensions (maybe other things) don't allow you to set `.innerHTML` for security reasons, so we "trick" it with this technique so it can't tell we're setting the innerHTML, lol

KubaJastrz: Sounds like perfect "security" implementation 😄

atomiks: I can't figure out how to trick minifiers (terser is smart enough to reduce it to .innerHTML) but for extensions people can use the unminified version

src/js/css.js
```diff
@@ -1,14 +1,10 @@
-import { isBrowserSupported } from './browser'
-
 /**
  * Injects a string of CSS styles to a style node in <head>
  * @param {String} css
  */
 export function injectCSS(css) {
-  if (isBrowserSupported) {
```
KubaJastrz: I think without this check it will break SSR when using `index.all.js`, because `document is not supported` error.

atomiks:
First alpha released: https://github.com/atomiks/tippyjs/releases/tag/v4.0.0-alpha.0

Play around & find issues 😗

KubaJastrz:
First issue coming right up: no `role` prop in types definition file.

atomiks:
That always gets out of sync. This is why we need to rewrite in TS which is automatic right?

KubaJastrz:
I haven't used it before, but there is a `--declaration` flag to TS compiler that generates `*.d.ts` files. I'm sure it can be done. I'll get on that when v4 stable is about to be shipped, so it can be rewritten as v4.1 perhaps.

atomiks:
Regarding the docs versioning, it seems that GitHub Pages can't have multiple branches, otherwise I'd keep v3 as a branch and have gh-pages published from there. 

I'm thinking of just linking the v4 release tag as the migration guide, & using the [this permalink](https://github.com/atomiks/tippyjs/tree/49b7dc9777c38d5da32a5e40df75579bfed6c757/website/src/pages) as the docs for the old version. They are relatively easy to read anyway.

website/src/components/Tippy.js
```diff
@@ -1,12 +1,139 @@
...
+function getNativeTippyProps(props) {
+  return Object.keys(props).reduce((acc, key) => {
+    if (REACT_ONLY_PROPS.indexOf(key) === -1) {
+      acc[key] = props[key]
+    }
+    return acc
+  }, {})
+}
+
+class Tippy extends React.Component {
```
KubaJastrz: I assume this is temporarily here until `@tippy.js/react@2` is out, right?

atomiks: Yup

atomiks:
Care for one final review? @KubaJastrz 

May release it tomorrow as long as everything is good, have been doing tons of manual testing for edge cases and added more unit tests + browser tests not covered by Jest

.babelrc
```diff
@@ -1,18 +1,11 @@
 {
-  "presets": [["env", { "modules": false }], "stage-2"],
+  "presets": ["@babel/env"],
+  "plugins": [
+    ["@babel/proposal-object-rest-spread", { "loose": true }, "unique"]
```
KubaJastrz: What does `"unique"` do here? I can't find anything about it in the babel docs

README.md
```diff
@@ -8,69 +8,24 @@
...
-      tippy.setDefaults({
-        arrow: true,
-        delay: 40,
-        theme: 'my-tippy',
-      })
-    </script>
-  </body>
-</html>
+# Yarn
+yard add tippy.js
```
KubaJastrz:
```suggestion
yarn add tippy.js
```

src/js/index.js
```diff
@@ -36,33 +36,15 @@ function tippy(targets, options, one) {
...
-    destroyAll() {
-      collection.instances.forEach(instance => {
-        instance.destroy()
-      })
-      collection.instances = []
-    },
-  }
-
-  return collection
+  return isSingular(targets) ? instances[0] : instances
```
KubaJastrz: Isn't that the other way around? If it is a singular then it can't have `0`-indexed item in it

src/js/popper.js
```diff
@@ -166,6 +165,18 @@ export function reflow(popper) {
   void popper.offsetHeight
 }
 
+/**
+ * Adds/removes theme from tooltip's classList
+ * @param {HTMLDivElement} tooltip
+ * @param {String} theme
+ * @param {String} action
+ */
+export function theme(tooltip, theme, action) {
```
KubaJastrz: Honestly, it was hard for me to decipher this function purpose without looking at the declaration here.
Maybe I've been using data-last approach too much recently but doesn't this `theme(tooltip, props.theme, 'add')` look kinda weird to you? :D

package.json
```diff
@@ -25,6 +28,7 @@
     "dev": "parcel demo/index.html -d .devserver --open",
     "build": "node rollup.build.js && bundlesize",
     "test": "jest --coverage",
+    "test:browser": "parcel test/browser/index.html -d .devserver --open",
```
KubaJastrz: We should probably implement automated visual tests, maybe cypress? I'll take a look at it

test/spec/popper.test.js
```diff
@@ -26,19 +26,8 @@ import {
...
-    tip.instances.forEach(i => i.show(0))
-    expect(document.querySelectorAll(Selectors.POPPER).length > 0).toBe(true)
-    hideAllPoppers()
-    setTimeout(() => {
-      expect(document.querySelectorAll(Selectors.POPPER).length).toBe(0)
-      done()
-    })
-  })
+describe('hideAll', () => {
+  // todo
```
KubaJastrz: Friendly reminder ^^

website/src/components/Nav.js
```diff
@@ -75,6 +77,25 @@ const XButton = styled.button`
...
+  margin-left: 20px;
+  margin-top: -10px;
+  margin-bottom: 10px;
+`
+
+function Hello() {
+  return <div app="name" style={{ hello: 'lol' }} />
+}
+
+Hello()
```
KubaJastrz: Debug code?

.babelrc
KubaJastrz: Ok, I found [this](https://babeljs.io/docs/en/options#pluginentry-presetentry), but that name link leads nowhere :/ So still no clue

src/js/popper.js
atomiks: Yeah I forgot about this and wanted to fix it as soon as I made it, it has `theme` in scope 3 different times lol

website/src/pages/themes.mdx
```diff
@@ -10,6 +10,45 @@ import Tippy from '../components/Tippy'
...
+</head>
+```
+
+**Package managers:**
+
+```js
+import 'tippy.js/dist/themes/light.css'
+import 'tippy.js/dist/themes/light-border.css'
+import 'tippy.js/dist/themes/google.css'
+import 'tippy.js/dist/themes/translucent.css'
```
KubaJastrz: This whole file still uses `tippy.js@3` and `dist/` directory

.babelrc
atomiks: Parcel was complaining about duplicate plugins, this fixed it

src/js/index.js
atomiks: It's always an array, if it's singular it's an array with one item in it, so we return it directly

KubaJastrz: `isSingular` checks whether the value is a plain object or instance of `Element`. Are you sure it is working as intended?

KubaJastrz: Ah sorry, for some reason I was thinking it was checking an array to be singular. My mistake

atomiks: Hah yes, all of the tests would be failing since it relies on `tippy()` returning correctly 😉

atomiks:
Thanks for the review, it's so helpful to have someone else look over all the mess of work/commits/testing/debugging you've been doing for the past couple weeks where it becomes easy forget certain things, etc. 🤕

atomiks:
Let's goooo ⚡️

mwjames:
> 4 formats:
-- index.js = tippy
-- index.min.js = tippy minified
-- index.all.js = tippy + css
-- index.all.min.js = tippy + css minified

Just last week we switched from qTip2 to tippy as tooltip base implementation but those file names in v4 (`index.js` ...) without a reference to tippy itself are somewhat hollow and require additional mingling to make them appear as being a tippy related JS resource especially when debugging. If possible, please retain the names or do you see popper using its own instances produced as index.js?

atomiks:
@mwjames The sourcemaps should allow for debugging

PR #424, opened 2019-02-10
Fix content as a function

atomiks:


index.d.ts
```diff
@@ -13,7 +13,7 @@ export type Placement =
   | 'right-start'
   | 'right-end'
 
-export type Content = string | Element
+export type Content = string | Element | ((ref: Element) => Element | string)
```
KubaJastrz: I don't think it will work with https://github.com/atomiks/tippyjs/blob/32002e849adc2480ad926e681466b144421a79f3/src/js/createTippy.js#L805-L810

atomiks: This functionality doesn't actually seem to be working in v4, I think it's a regression? Think we neglected it, evidenced by the lack of typing

atomiks: Test with `instance.set()` appears to work with this new change.

KubaJastrz: No, I meant `content` as a function wouldn't have been called. Good job

PR #427, opened 2019-02-14
Fix UC Browser support

atomiks:
So had some "fun" debugging why it wasn't working on the great UC Browser through Browserstack 😒 

- `animateFill` effect doesn't work, probably a rendering glitch. The tooltip text is unreadable because white text on white background without the backdrop. Let's disable it if we detect UC Browser
- `transitionend` does not fire because it uses `webkitTransitionEnd` _despite_ supporting unprefixed `transitions` (in the CSS, only transforms use prefixes. Unprefixed transitions have wider support). I can only hope they don't remove `webkitTransitionEnd` 

- Also a check in case the `<head>` doesn't have any nodes in it, we'll use `.appendChild` instead of inserting it before the first node

### DIFF
--- a/src/js/css.js
+++ b/src/js/css.js
@@ -9,6 +9,12 @@ export function injectCSS(css) {
     const style = document.createElement('style')
     style.type = 'text/css'
     style.textContent = css
-    document.head.insertBefore(style, document.head.firstChild)
+    const head = document.head
+    const { firstChild } = head
+    if (firstChild) {
+      head.insertBefore(style, firstChild)
+    } else {
+      head.appendChild(style)
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


src/js/css.js
```diff
@@ -9,6 +9,12 @@ export function injectCSS(css) {
     const style = document.createElement('style')
     style.type = 'text/css'
     style.textContent = css
-    document.head.insertBefore(style, document.head.firstChild)
+    const head = document.head
+    const { firstChild } = head
+    if (firstChild) {
+      head.insertBefore(style, firstChild)
+    } else {
+      head.appendChild(style)
```
KubaJastrz: Why not just append in all cases?

atomiks: We want to append to the top of the `<head>` so styles can be overridden by the user's stylesheets below it

atomiks:
I ended up adding `&& document.body.style.WebkitTransition !== undefined` as well for extra safety, if they do ever remove webkit transitions in the future.

Issue #408, opened 2019-01-14
export 'Props' was not found in 'tippy.js'


nadavshabtai:
Hi

I'm using tippy 3.4.1 and I'm having trouble importing tippy Props and Instace using typescript because I get the following warning:

```
WARNING in [path]
"export 'Props' was not found in 'tippy.js'
[stack trace]
```
am I importing it wrong?
`import tippy, {Instance, Props} from 'tippy.js';`

KubaJastrz:
This should work, can you post your `tsconfig.json`?

nadavshabtai:
this is my tsconfig.json
```
{
  "compilerOptions": {
    "baseUrl": "",
    "declaration": false,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "lib": ["es2015", "es2016.array.include", "dom"],
    "mapRoot": "./",
    "module": "es2015",
    "moduleResolution": "node",
    "sourceMap": true,
    "target": "es5",
    "typeRoots": [
      "node_modules/@types"
    ]
  },
  "exclude": [
    "node_modules",
    "src/main-prod.ts"
  ]
}
```
I managed to solve this problem using the following code in d.ts file, but I don't think it's pretty... 

```
declare module 'tippy.js/dist/tippy.standalone' {
    import { Tippy } from 'tippy.js';
    export { Instance, Props, BasicPlacement, Placement, Content, Targets, Collection } from 'tippy.js';
    const tippy: Tippy;
    export default tippy;
}
```

KubaJastrz:
I can't reproduce this. So you're importing tippy via `tippy.js/dist/tippy.standalone` now?

Btw, you can also do `export * from 'tippy.js`.

nadavshabtai:
1. Yes, because of the 'module' in the package.json it's using standalone, but I import from tippy.js just as I wrote on the top.

2. You're right :) stupid me

atomiks:
Is this a problem in v4?

atomiks:
If you can provide a reproduction repo I'll repon!

PR #441, opened 2019-03-08
TypeScript rewrite

### LINKED ISSUE #378: Convert codebase to TypeScript
@KubaJastrz 

I've finished adding JSDoc comments for now, but it would be cool to have the codebase in TypeScript instead (as I've put off learning it for too long now 😞). Would you be interested in helping out?

atomiks:
Resolves #378 

cc @KubaJastrz, also how do I use a single declaration file (`./types.ts` and `index.d.ts` are duplicates?)


KubaJastrz:
Definitely a good start. There are still many `as` and `any` keywords that could be replaced. And `.eslintrc` is not set up correctly for typescript plugin.

I'm gonna do a full review later today locally and check what can be written more strictly.

atomiks:
One thing that really confuses me is the angle bracket syntax `<>`. I don't understand it at all. I read the docs for it and it's meant to be like `as`(?) but I don't get it, yet see it everywhere 🤯

KubaJastrz:
They can be used for type casting, but only in `.ts` files (that is not the case for `.tsx` in React), so I recommend not using it this way at all.

Another use that you might see (basically all the time) is as generic type. It works sort of like a function parameter. More info [here](https://www.typescriptlang.org/docs/handbook/generics.html).

atomiks:
Yeah I thought the generic type would be good for the util functions but seemed to cause problems so just fell back to `any` :\. Anyway look forward to your improvements 🙂

src/index.ts
```diff
@@ -54,9 +56,9 @@ tippy.defaults = Defaults
 /**
  * Static methods
  */
-tippy.setDefaults = partialDefaults => {
+tippy.setDefaults = (partialDefaults: Props) => {
```
KubaJastrz:
```suggestion
tippy.setDefaults = (partialDefaults: Partial<Props>) => {
```

src/index.ts
```diff
@@ -54,9 +56,9 @@ tippy.defaults = Defaults
...
   Object.keys(partialDefaults).forEach(key => {
-    Defaults[key] = partialDefaults[key]
+    ;(Defaults as any)[key] = (partialDefaults as any)[key]
```
KubaJastrz: You can add `// @ts-ignore` until we find a solution instead.

src/ponyfills.ts
```diff
@@ -1,6 +1,6 @@
 import { isBrowser } from './browser'
 
-const elementProto = isBrowser ? Element.prototype : {}
+const elementProto = isBrowser ? Element.prototype : ({} as any)
```
KubaJastrz:
```suggestion
const elementProto: Record<string, any> = isBrowser ? Element.prototype : {}
```

`Record<TKey, TValue>` evaluates to `{ [key: TKey]: TValue }` and is common to describe an actual object (not non-primitive, like `object`).

src/utils.ts
```diff
@@ -78,69 +68,53 @@ export function debounce(fn, ms) {
...
   return (
     !!(value && hasOwnProperty(value, 'isVirtual')) || value instanceof Element
   )
 }
 
 /**
  * Firefox extensions don't allow setting .innerHTML directly, this will trick it
- * @return {String}
  */
 export function innerHTML() {
```
KubaJastrz:
```suggestion
export function innerHTML(): 'innerHTML' {
```

This will fix those ugly `;(element as any)[innerHTML()]` things.

src/popper.ts
```diff
@@ -1,37 +1,41 @@
...
- * @param {Object} props
  */
-export function setContent(contentEl, props) {
+export function setContent(contentEl: PopperChildren['content'], props: Props) {
   if (props.content instanceof Element) {
     setInnerHTML(contentEl, '')
     contentEl.appendChild(props.content)
   } else {
+    // @ts-ignore
     contentEl[props.allowHTML ? 'innerHTML' : 'textContent'] = props.content
```
KubaJastrz: The warning is correct. What if `props.content` is a function?

To avoid unknown index, you can use
```ts
const key: 'innerHTML' | 'textContent' = props.allowHTML ? 'innerHTML' : 'textContent'
contentEl[key] = props.content
```

src/popper.ts
```diff
@@ -123,36 +122,36 @@ export function applyTransitionDuration(els, value) {
...
+  listener: EventListener,
+) {
   // UC Browser hasn't adopted the `transitionend` event despite supporting
   // unprefixed transitions...
   const eventName =
     isUCBrowser && document.body.style.webkitTransition !== undefined
       ? 'webkitTransitionEnd'
       : 'transitionend'
-  tooltip[action + 'EventListener'](eventName, listener)
+  ;(tooltip as any)[action + 'EventListener'](eventName, listener)
```
KubaJastrz:
```suggestion
  tooltip[(action + 'EventListener' as 'addEventListener' | 'removeEventListener')](eventName, listener)
```

KubaJastrz: Or extract to `key` variable as above.

src/reference.ts
```diff
@@ -34,41 +33,42 @@ export function getDataAttributeOptions(reference) {
 /**
  * Polyfills the virtual reference (plain object) with Element.prototype props
  * Mutating because DOM elements are mutated, adds `_tippy` property
- * @param {Object} virtualReference
  */
-export function polyfillElementPrototypeProperties(virtualReference) {
-  const polyfills = {
+export function polyfillElementPrototypeProperties(
+  virtualReference: VirtualReference,
```
KubaJastrz:
```suggestion
  virtualReference: VirtualReference & Record<string, any>,
```

I don't know if this is the best way, but it will help with `;(virtualReference as any)[key] = polyfills[key]`. Also `const polyfills: Record<string, any>`

src/createTippy.ts
```diff
@@ -65,10 +72,10 @@ export default function createTippy(reference, collectionProps) {
   let isScheduledToShow = false
 
   // The current `transitionend` callback reference
-  let transitionEndListener = () => {}
+  let transitionEndListener: EventListener
 
   // Array of event listeners currently attached to the reference element
-  let listeners = []
+  let listeners: object[] = []
```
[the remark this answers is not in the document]
KubaJastrz: If you want an actual object, you can use `Record<string, any>`. I've described what it does somewhere below.

src/createTippy.ts
```diff
@@ -215,7 +219,7 @@ export default function createTippy(reference, collectionProps) {
       }),
       clientWidth: 0,
       clientHeight: 0,
-    }
+    } as ReferenceElement
```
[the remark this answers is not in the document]
atomiks: It's actually not a bug because Popper only uses those three properties (and we don't override the Tippy instance's reference, only the Popper's. But still, I'll merge the previous object in 😄

.eslintrc
```diff
@@ -8,5 +9,8 @@
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": 2018
+  },
```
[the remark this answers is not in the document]
atomiks: Seemed to throw errors after that change.

atomiks:
Thank you tons for the help @KubaJastrz ! I think we've caught a few obscure bugs thanks to TS 😎 

It seems like there are lots of these errors now:

> [ts] Cannot invoke an object which is possibly 'undefined'. [2722]

> [ts] Object is possibly 'null'. [2531]

> [ts] Object is possibly 'undefined'

With regards to the instance props, etc.  Any ideas of what to do?

atomiks:
I think we need to distinguish the Props interface and Options

Like this?

```js
export interface Props {
  [all required, no ?:]
}

export type Options = Partial<Props>
```

That would probably be a breaking change for the types though...? `:/`

atomiks:
Managed to get all types passing 🤷‍♂️ 

Had to hack the `EventListener` type for the on* trigger listeners. I'm sure there's a better way

src/index.ts
```diff
@@ -31,16 +29,20 @@ function tippy(targets, options) {
...
   }
 
-  const instances = getArrayOfElements(targets).reduce((acc, reference) => {
-    const instance = reference && createTippy(reference, props)
-    if (instance) {
-      acc.push(instance)
-    }
-    return acc
-  }, [])
+  const instances = getArrayOfElements(targets).reduce(
```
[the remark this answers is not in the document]
atomiks: Not sure how to make this one work.

src/createTippy.ts
```diff
@@ -540,15 +533,20 @@ export default function createTippy(reference, collectionProps) {
...
         }
       },
-    })
+    }
+
+    instance.popperInstance = new Popper(
+      instance.reference,
+      instance.popper,
+      config,
+    ) as PopperInstance
```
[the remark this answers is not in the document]
atomiks: Yes, it seems like the Popper namespace doesn't have the correct properties. Have to create our own?

.eslintrc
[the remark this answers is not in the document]
KubaJastrz: True, I'm not sure why though. It was working for me before iirc.

src/createTippy.ts
```diff
@@ -830,9 +822,9 @@ export default function createTippy(
       ...options,
       ignoreAttributes: true,
     })
-    nextProps.ignoreAttributes = hasOwnProperty(options, 'ignoreAttributes')
+    nextProps.ignoreAttributes = (hasOwnProperty(options, 'ignoreAttributes')
       ? options.ignoreAttributes
-      : prevProps.ignoreAttributes
+      : prevProps.ignoreAttributes) as boolean
```
KubaJastrz: I think `hasOwnProperty` could also check for `undefined`, so it falls back to `prevProps` in this case.

```ts
const options = { ignoreAttributes: undefined }
hasOwnProperty(options, 'ignoreAttributes') // true
```

If it was `false` then we could write `options.ignoreAttributes!` in the ternary and skip `as boolean` assignment.

src/createTippy.ts
```diff
@@ -845,7 +837,10 @@ export default function createTippy(
 
     if (hasOwnProperty(options, 'interactiveDebounce')) {
       cleanupOldMouseListeners()
-      debouncedOnMouseMove = debounce(onMouseMove, options.interactiveDebounce)
+      debouncedOnMouseMove = debounce(
+        onMouseMove,
+        options.interactiveDebounce as number,
```
KubaJastrz: Same here, `options.interactiveDebounce` could be `undefined`, that's why TS throws an error.

src/popper.ts
```diff
@@ -259,15 +266,15 @@ export function updatePopperElement(
     tooltip.appendChild(createBackdropElement())
     tooltip.setAttribute('data-animatefill', '')
   } else if (prevProps.animateFill && !nextProps.animateFill) {
-    tooltip.removeChild(backdrop)
+    tooltip.removeChild(backdrop as Node)
```
KubaJastrz: Use `backdrop!` as we know it's not null.

src/popper.ts
```diff
@@ -259,15 +266,15 @@ export function updatePopperElement(
...
     tooltip.removeAttribute('data-animatefill')
   }
 
   // arrow
   if (!prevProps.arrow && nextProps.arrow) {
     tooltip.appendChild(createArrowElement(nextProps.arrowType))
   } else if (prevProps.arrow && !nextProps.arrow) {
-    tooltip.removeChild(arrow)
+    tooltip.removeChild(arrow as Node)
```
KubaJastrz: Same here, `arrow!`.

src/popper.ts
```diff
@@ -276,7 +283,7 @@ export function updatePopperElement(
     nextProps.arrow &&
     prevProps.arrowType !== nextProps.arrowType
   ) {
-    tooltip.replaceChild(createArrowElement(nextProps.arrowType), arrow)
+    tooltip.replaceChild(createArrowElement(nextProps.arrowType), arrow as Node)
```
KubaJastrz: And here. We have checks in those places.

tsconfig.json
```diff
@@ -3,7 +3,9 @@
     "moduleResolution": "node",
     "noEmit": true,
     "resolveJsonModule": true,
-    "noImplicitAny": true
+    "noImplicitAny": true,
```
KubaJastrz: `"noImplicitAny": true` can be dropped because `"strict": true` already includes it.

src/createTippy.ts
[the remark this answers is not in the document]
KubaJastrz: If they are not needed, then probably yes.

.eslintrc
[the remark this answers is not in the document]
KubaJastrz: You can check [my fork](https://github.com/KubaJastrz/tippyjs/blob/29a5ffc24b083911a25406b100268794e30e05de/.eslintrc) (my outdated attempt to introduce TS :P).

Also, I have the plugin working just fine.

atomiks:
> We could drop index.d.ts and replace it with --declaration flag during the build process. I'll look into it.

We need to keep the `Props` backwards-compat though. So I think we should keep it separate for this major?

KubaJastrz:
There's something weird going on. We have our internal types in `src/types.ts`. That's where I proposed to create `Defaults` interface that just mirrors the object in `src/defaults.ts`.

So the flow would be something like this:
```ts
// in src/types.ts
import defaults from './defaults'
type Defaults = typeof defaults
interface Props extends Defaults /* maybe omit some props */ {
  /* and override them here, like we do in @tippy.js/react */
}
```
No need for `Options` interface.

Because we're using typescript, we could potentially drop `index.d.ts` and have it be generated automatically during build process (I hope this can be done without breaking changes, not sure).

atomiks:
Well ideally we don't have a Defaults interface. I only added it as part of `index.d.ts` backwards compat. We should rename that file I guess and move props.ts stuff into utils.ts.

There should be two interfaces: `Props` and `Options = Partial<Props>`

atomiks:
@KubaJastrz I think this is pretty decent now? I'll keep `index.d.ts` and `./types.ts` separate for now, we can work out what to do in next major. Currently they're synced with some internal/private types taken away.

KubaJastrz:
> Well ideally we don't have a Defaults interface.

So why do we have it now? I thought that could be used for [fixing this](https://github.com/atomiks/tippyjs/pull/441#discussion_r263866819), but currently it's just a copy of previous `Props` interface.

atomiks:
> So why do we have it now? I thought that could be used for fixing this, but currently it's just a copy of previous Props interface (without ?).

Cuz I copy-pasted the Props interface from `./types` which included a couple fixes 😛 Will be easier to keep them in sync if I don't need to convert them to be optional in `index.d.ts`

atomiks:
Well `hideAll` includes an internal check that makes it harder to for users to implement. `group` is possible to implement outside but figured it would be better built-in since it's a little complex.

I'm going to avoid adding (easy) features to core if they can be implemented by users without much trouble, to avoid increasing bundle size. Thanks for the request though 🙂

atomiks:
Going to merge this in now as linting, types and tests pass fine, I'll do a minor release in the next week sometime. Feel free to make some PRs in the meantime to improve the current types 🙂

KubaJastrz:
I don't know what `WholeProps` would be useful for. I think it's unnecessary because you can easily use something like [this](https://github.com/Microsoft/TypeScript/issues/15012#issuecomment-363399211) in the userland :
```ts
type WholeProps = Required<Options>
```

I suggest just sticking to `Props` -> `Options` rename if you really want it.

atomiks:
It types `instance.props` (default props + user-supplied options), but other interface use `Options` (only user-supplied options).

People were doing https://github.com/atomiks/tippyjs/issues/408;

```js
import { Props } from 'tippy.js'
```

But it will break if we change `Props` to be `Required<Options>` (or use non-optional interface)?

PR #443, opened 2019-03-10
Remove WholeProps from index.d.ts

KubaJastrz:
Ref: https://github.com/atomiks/tippyjs/pull/441#issuecomment-471278734

---

You can use custom `Required<T>` type as suggested [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#improved-control-over-mapped-type-modifiers).

```ts
import {Options} from 'tippy.js'

type Props = Required<Options>
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,66 +43,60 @@ export interface PopperInstance extends Popper {
   modifiers: { name: string; padding: object | number }[]
 }
 
-/**
- * This interface will be called `Props` in the next major.
- * Exists for backwards-compat after TypeScript rewrite.
- */
-export interface WholeProps {
-  a11y: boolean
-  allowHTML: boolean
-  animateFill: boolean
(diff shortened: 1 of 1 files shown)


atomiks:
This is def a better solution I think 👍 

`WholeProps` should be `Props` though like in `src/types.ts`, they'll just avoid the exported `Props` (deprecated). In v5 it will match the internal types in src?

KubaJastrz:
In v5 `Props` will be removed from user access. Only `Options` will be available (that functions exactly the same as `Props` does now).

The problem is that our internal `Props` are slightly misleading currently. They are meant to represent the type of `defaults`, but they don't (they include union types, it should be exact 1:1 match).

atomiks:
Does `defaultProps` need an exact type? `Props` was meant to represent the merged result of the default props + user options throughout the code (instance.props)

I'll let you handle any fixes as PRs 😝

KubaJastrz:
Okay, maybe it will be easier to explain my thoughts in code :p

PR #444, opened 2019-03-10
Improve utils typings, fix setTimeout

KubaJastrz:
Simplifies `getArrayOfElements` types and fixes weird issue with `setTimeout`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-import { Props, ReferenceElement, VirtualReference } from './types'
+import { Props, ReferenceElement, Targets, VirtualReference } from './types'
 import { arrayFrom, matches } from './ponyfills'
 import { isUCBrowser } from './browser'
 import { getDataAttributeOptions } from './reference'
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "noEmit": true,
     "resolveJsonModule": true,
     "strict": true,
-    "noUnusedLocals": true
+    "noUnusedLocals": true,
+    "types": []
```
KubaJastrz: TypeScript was loading `@types/node` for some reason and that made `setTimeout` go crazy. We don't have any dependencies with `@types/...` so its empty for now.

src/utils.ts
```diff
@@ -86,7 +87,7 @@ export function includes(a: any[] | string, b: any): boolean {
 /**
  * Determines if the value is singular-like
  */
-export function isSingular(value: any): boolean {
+export function isSingular(value: any): value is VirtualReference | Element {
```
KubaJastrz: Ref: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

atomiks:
Awesome 😎

PR #445, opened 2019-03-11
Use microbundle standard packaging

atomiks:
This doesn't switch to microbundle, but uses [microbundle](https://github.com/developit/microbundle) packaging fields as suggested in README. 

Adds `unpkg` field. The code in the `main` field should be unminified.

Note: It doesn't use the same directory format, microbundle uses `dist/*` with format file names. We're using directories here because there are a bunch of different formats and it's easier to import this way.

`import tippy from 'tippy.js/esm'` => import without CSS

Instead of:

`import tippy from 'tippy.js/dist/tippy.mjs'` => 😕


atomiks:
I wonder if we should also run prettier over the dist files, because the output isn't very pretty. Although sourcemaps kind of alleviate that anyway

atomiks:
I wonder if we should also support a "modern" package that uses ES2018 (?) syntax, for smaller bundle sizes for those who don't need to support legacy browsers 🤔 

They could do:

```js
import tippy from 'tippy.js/modern'
```

rollup.build.js
```diff
@@ -63,7 +63,7 @@ const configs = {
 }
 
 const outputs = {
-  bundle: [output('umd'), output('esm')],
+  bundle: [output('umd'), output('cjs'), output('esm')],
```
KubaJastrz: Currently, all bundles are the same, just in different directories, right?

KubaJastrz:
> Also I think we need to update the `index.d.ts` to declare these modules

It can be done with `declare module`.

rollup.build.js
atomiks: Yep: `index.all.js` + `index.js` + minified versions of those + sourcemaps

atomiks:
> It can be done with declare module.

@KubaJastrz can you push to this branch with this?

atomiks:
After thinking about it more, going to remove the `cjs` bundle. UMD works the same as CJS in Node envs anyway with a few extra bytes. But I'd still like `"unpkg"` field and unminified code in `"main"`.

Also we need to support browser-only `esm`, because `import Popper from 'popper.js'` does not work in a browser. But I'd like to avoid bundling it :\

atomiks:
## Browser ESM version

I'm thinking we can have:

`./esm/index.all.browser.js` (+ minified)

No bundling of Popper.js. Some post-rollup build step can just replace this line:

```diff
- import Popper from 'popper.js'
+ import Popper from 'https://unpkg.com/popper.js@1/dist/esm/popper.min.js'
```

Same extra HTTP request as the UMD version. And users can still import Popper.js themselves without extra size. And it won't become outdated. 

## Modern version (ES2017)

`./esm/index.all.modern.js` (+ minified)

For users to place in bundles that use `type="module"`. This will just use a custom browserslist config, and CSS without any `-webkit-` prefix transforms.

atomiks:
Okay the "modern mode" is pointless because it doesn't keep the default params/destructuring (looking around, seems like Edge didn't support it for a while). So it ended up being like 5 bytes larger 🤦‍♂️ 

I've added a browser ESM version though

atomiks:
Turns out unpkg already does browser ESM properly, although it's not minified. Seems to be an open thread on it though. Gonna close this out... 😪

Issue #448, opened 2019-03-11
Placement auto removed ? 


miguelyoobic95:
I was previous using the `placement: "auto"` option in my code and it has suddenly stopped working after upgrading to the latest version of the library.

Is this functionality still available or is there a new way to implement the use case where the tooltip itself should work out the correct position ?

Thanks in advance. Best

Miguel

atomiks:
@KubaJastrz I guess we're missing a type for `"auto"`
We should import the types from popper.js

PR #449, opened 2019-03-11
Use imported Placement

KubaJastrz:
Import `Placement` from `popper.js` (https://github.com/atomiks/tippyjs/issues/448#issuecomment-471646380).


KubaJastrz:
@atomiks, what about this? https://github.com/atomiks/tippyjs/blob/d08e8a10dea7b5faf7707247c36a2188345022d0/src/popper.ts#L152-L158

atomiks:
`"auto"` gets converted to one of the real placements, so it should be fine

PR #450, opened 2019-03-13
Add x-placement to tooltip node

atomiks:
No need to think about `.tippy-popper` in themes anymore.

Don't know why I didn't think of doing this sooner 🤦‍♂️

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -21,24 +21,15 @@
         background: #eee;
         border: 1px solid #ddd;
         font-size: 16px;
-        margin: 8px;
+        margin: 20px;
         padding: 8px;
       }
     </style>
   </head>
 
   <body>
-    <div class="container">
-      <button data-tippy="Auto">Auto</button>
-
-      <h2>Group</h2>
-      <button class="group">Group</button>
-      <button class="group">Group</button>
-      <button class="group">Group</button>
-      <button class="group">Group</button>
-
-      <h2>Event delegation</h2>
-      <button class="delegation" data-tippy-content="Event delegation"></button>
-    </div>
+    <div class="container"></div>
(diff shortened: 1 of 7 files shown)


demo/index.html
```diff
@@ -21,24 +21,15 @@
...
-      <button class="group">Group</button>
-      <button class="group">Group</button>
-      <button class="group">Group</button>
-
-      <h2>Event delegation</h2>
-      <button class="delegation" data-tippy-content="Event delegation"></button>
-    </div>
+    <div class="container"></div>
+    <!-- prettier-ignore -->
+    <script> </script>
```
KubaJastrz: ?

atomiks: Fixes a Chrome transition bug that's existed for years. I'm using `showOnInit` in the demo and the tooltips transition weirdly on load without this hack. A script tag with an empty space fixes it (so weird, I know)

atomiks: https://stackoverflow.com/a/42969257/9470278

KubaJastrz: Huh, interesting. Please, add a comment about this for the future generations 😉.

atomiks:
Doesn't update the attribute when you do

`instance.set({ placement: new })`

🤔

atomiks:
Looks like this exposed a bug when updating a popperInstance dependency: it should be calling `onCreate` too (reason why it doesn't have to usually is because we're setting `onCreate` and `onUpdate` in `afterPopperPositionUpdates`.

atomiks:
That `afterPopperPositionUpdates` hack only exists due to Popper's lack of callback / promise return type for `.update() / scheduleUpdate()` too... 😞 

I think this should be good now.

PR #454, opened 2019-03-24
Change notransition technique

atomiks:
Fixes https://github.com/atomiks/tippy.js-react/issues/67

Changes the "notransition" technique which fixes undesirable transitions when the tippy changes placements on updates.

**Before**: add a `transition: none` CSS class on the tippy once `onShown()` completes

**After**: use some instance flags to check if the previous placement is the same and if the previous update happened while the tippy was visible. If so, disable transitions for a single frame before painting. This is a better way to do it anyway and it fixes the problem linked

(going to be deleting the comments describing the variables after this as they're quite redundant, that's why I didn't add any here)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -7,7 +7,6 @@ export const PADDING = 4
 // Classes
 export const IOS_CLASS = 'tippy-iOS'
 export const ACTIVE_CLASS = 'tippy-active'
-export const NO_TRANSITION_CLASS = 'tippy-notransition'
 
 // Selectors
 export const POPPER_SELECTOR = '.tippy-popper'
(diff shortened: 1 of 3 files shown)


Lazyuki:
@atomiks Thanks for the quick fix! I assume this will be included in `tippy.js@4.2.1`, but since `tippy.js-react` depends on `tippy.js^4.0.4` right now, could you update it to `tippy.js^4.2.0`?

atomiks:
@Lazyuki that semver range means when you update `@tippy.js/react` you'll get the latest `tippy.js` version anyway.

Run

```bash
npm i @tippy.js/react@latest
```

in your project. I'm going to release this patch today

PR #456, opened 2019-03-24
Optimize set(): avoid recreating popperInstance by comparing props

atomiks:
cc @KubaJastrz not sure if this type can be better

This will compare the previous props with the new options being passed to `set()` - if none of them changed, there is no point recreating the `popperInstance` (which is somewhat expensive). 

This is mainly for the React wrapper which has no diffing (to conserve bundle size) because the props will always be passed to `set()` on each render, leading to the `popperInstance` being recreated each time. In the vanilla lib usually they won't be doing this [I think]

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
       { "functions": false }
     ],
     "@typescript-eslint/no-non-null-assertion": "off",
-    "@typescript-eslint/no-explicit-any": "off"
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/array-type": ["off"]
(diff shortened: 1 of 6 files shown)


KubaJastrz:
Fixed types in 8864d9c.

About [eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md): to comply with the `"array"` setting, I'd have to write `(keyof Props)[]` which I find less readable, but I can revert it if you want.

atomiks:
Great, thanks!

PR #458, opened 2019-03-27
Linting

brettz9:
- Linting (ESLint): Add `.eslintignore` (Useful for IDE to avoid flagging when browsing files)
- Linting (ESLint): Apply ESLint to `rollup.build.js`
- Linting (ESLint): Fix error of missing `done` in test file (though `test` is in the ignore file to avoid flagging lesser errors)
- Linting (npm): Add linter-recommended `contributors`, `bugs`, `homepage`, and `engines` properties (some of these used on npmjs.com)
- Linting (HTML): Avoid favicon request in console

As far as the `test` linting errors, I personally think it would be better to ensure you are catching any errors there (if not in `website` too), but as it didn't appear to be your intent to lint them, I've added `test` and `website` to `.eslintignore` as well as the `dist` paths.

See above for the other rationales of inclusion.

I plan to provide a more substantive PR for modules after this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,13 @@
   "unpkg": "./umd/index.all.min.js",
   "types": "index.d.ts",
   "author": "atomiks",
+  "contributors": [
+    "Brett Zamir"
+  ],
   "license": "MIT",
+  "bugs": "https://github.com/atomiks/tippyjs/issues",
+  "homepage": "https://atomiks.github.io/tippyjs/",
+  "engines": {},
   "keywords": [
     "tooltip",
     "popover",
@@ -31,7 +37,7 @@
     "test": "jest --coverage",
     "test:browser": "parcel test/browser/index.html -d .devserver --open",
     "check-types": "tsc",
-    "lint": "eslint \"src/**/*.ts\"",
+    "lint": "eslint --report-unused-disable-directives .",
     "clean": "rimraf umd esm themes coverage .devserver .cache ./index.css",
     "prepare": "npm run clean && npm run build"
   },
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   "license": "MIT",
   "bugs": "https://github.com/atomiks/tippyjs/issues",
   "homepage": "https://atomiks.github.io/tippyjs/",
-  "engines": {},
   "keywords": [
     "tooltip",
     "popover",
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+node_modules
+coverage
+esm
+umd
+website
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,6 +16,17 @@
   "parserOptions": {
     "project": "./tsconfig.json"
   },
+  "overrides": [{
+    "files": ["website/**"],
+    "rules": {
+      "@typescript-eslint/camelcase": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/no-use-before-define": "off",
+      "@typescript-eslint/no-var-requires": "off",
+      "react/prop-types": "off",
+      "react/no-unescaped-entities": "off"
+    }
+  }],
   "rules": {
     "@typescript-eslint/no-use-before-define": [
       "error",
@@ -23,6 +34,6 @@
     ],
     "@typescript-eslint/no-non-null-assertion": "off",
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/array-type": ["off"]
+    "@typescript-eslint/array-type": "off"
   }
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,28 +1,25 @@
-{
-  "env": {
-    "browser": true,
-    "node": true,
-    "jest": true,
-    "es6": true
+module.exports = {
+  env: {
+    browser: true,
+    node: true,
+    jest: true,
+    es6: true,
   },
-  "extends": [
-    "eslint:recommended",
-    "plugin:@typescript-eslint/recommended",
-    "prettier",
-    "prettier/@typescript-eslint"
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'prettier',
+    'prettier/@typescript-eslint',
   ],
-  "plugins": ["@typescript-eslint"],
-  "parser": "@typescript-eslint/parser",
-  "parserOptions": {
-    "project": "./tsconfig.json"
+  plugins: ['@typescript-eslint'],
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: './tsconfig.json',
   },
-  "rules": {
-    "@typescript-eslint/no-use-before-define": [
-      "error",
-      { "functions": false }
-    ],
-    "@typescript-eslint/no-non-null-assertion": "off",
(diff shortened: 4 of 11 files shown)


brettz9:
I just don't like getting error messages which are not actually errors distracting me while browsing files, and I figured that it may be preferable to use a forward-looking syntax rather than disabling the `require`-complaining ESLint rule (while still getting the benefits of ESLint for that file rather than ignoring it).

Re: Typescript/ts-node, I tried but it won't handle async and possibly other features. While there are other workarounds, I figure it is more trouble than it is worth. So if you prefer, I can just revert the `import`'s in that file to `require` and add an additional ESLint disable directive to ignore the complaint.

KubaJastrz:
Converting completely valid code just because linters don't like it (they are not even set up to check this file) is not desired. And `require` is node built-in to handle imports, so there's no reason to change it.

Regarding ts-node, I don't remember encountering those issues before. I'll see if it can be done.

brettz9:
Re: linters not set up to check the file, I understand, but my point is that when I use an IDE to look at the file, it is not aware of your linting script. It can only check `.eslintignore` (and besides, I didn't think there was anything wrong with checking an extra file to avoid mistakes--such as I caught in one of your testing files). But yeah, I understand about `require`.

I reverted to a version using `require` before I saw your comment, but I kept the `import` version, so I can easily go back to it.

brettz9:
Regarding getting it to work with ts-node, there were some suggestions in https://github.com/TypeStrong/ts-node/issues/51

package.json
```diff
@@ -27,7 +33,7 @@
   },
   "scripts": {
     "dev": "parcel demo/index.html -d .devserver --open",
-    "build": "node rollup.build.js && bundlesize",
+    "build": "npx babel-node rollup.build.js && bundlesize",
```
KubaJastrz: You don't need to transpile it with babel then.

brettz9: Whoops, fixed (including the `@babel/node` dep. I added).

brettz9:
Ok, I've made the changes so this PR should be ready if ok with you.

brettz9:
Regarding "(including use of modules)" that was an outdated comment which I've removed. But it is applying ESLint now to the Rollup file in the sense that the `.eslintignore` file isn't ignoring the file and it is passing linting tests (though I disabled two rules at the top of the file to do so--`no-console` and `@typescript-eslint/no-var-requires`).

I have not really used TypeScript--mostly have just peacefully co-existed with it through type files in repos I've used. :)

atomiks:
The linting script is `"eslint "src/**/*.ts"` in package.json, currently this PR doesn't lint the Rollup build file?

brettz9:
Right, I wasn't sure if you wanted it to lint that file or not, but I at least make ESLint-related fixes to the rollup file (and kept it out of `.eslintignore`). I can add that to the `eslint "src/**/*.ts` path or just change it to `eslint .` now that `.eslintignore` has the paths we don't want to be checked (and if you decide you do want the tests linted too, I can remove the `test` directory from the ignore file too).

atomiks:
I guess we can try to lint as much as possible

brettz9:
Ok, I've applied ESLint to the whole repo, excepting `esm`/`umd` and `coverage` (I had neglected to add the latter in the ignore file before).

For `website/`, I also added an ESLint React plugin so that in linting your website code, it would be aware of globals and such; note that it could not auto-detect the React version since React wasn't a direct dev dependency; let me know if you'd like me to add it as a dev dep so we don't need to hard-code the React version in the ESLint settings. Though this plugin also adds some rules you might want, for now, and for `website/` only, I have disabled the ones that were producing warnings, as well as disabling some typescript warnings.

I also added a `globals` directive to the Prism file there and fixed one linting complaint (unnecessary regex escape).

For the ESLint npm script, I added the flag `--report-unused-disable-directives` as I think it is nice to ensure one doesn't have disable directives lying around that one doesn't need.

And for the testing files, I made some additional linting related fixes, and in one case, temporarily disabled a rule (for sparse arrays).

Once this may be ok, I can then rebase my other PR.

brettz9:
Ok, I've removed the plug-in and am ignoring the website directory.

KubaJastrz:
It is still good to lint the website directory, but let's have separately in its own [`package.json`](https://github.com/atomiks/tippyjs/blob/0a824c6d7ec13fd85a4c886cc883e3945766998d/website/package.json).

brettz9:
Why don't I let you do that, as I don't know the combo of ESLinting you want for that...

KubaJastrz:
Of course, great job on this.

atomiks:
Thank you for the work @brettz9 ! We can do that in a different PR

I've made some updates to the files should be ready to merge now. Is the`"overrides"` in `.eslintrc` necessary now? Since we'll be making a new configuration for the website?

brettz9:
No, you're right, it isn't... I've remove it...

brettz9:
And FYI on the React plugin, the npm package was `eslint-plugin-react`, and the settings I added are at https://github.com/brettz9/tippyjs/commit/8039b8a142f30dab6dc20c5fc70226c1aef92980#diff-1dc6ee56b778cd91e0327b52aaeaa1b9 . You shouldn't need `settings` there because `react` is within *that* `package.json` so the plug-in can auto-detect the version if you are installing within `website/`. Some of the rules I had there in `overrides` might still be necessary to address, though now they won't be as `overrides`, they'll just be the rules in that directory's `.eslintrc.js`. (By the way, adding the `.js` extension is now recommended with the old way [deprecated](https://eslint.org/docs/user-guide/configuring#configuration-file-formats), so you may want to add the extension in both cases too.)

atomiks:
> (By the way, adding the .js extension is now recommended with the old way deprecated, so you may want to add the extension in both cases too.)

Should we update `.eslintrc` -> `.eslintrc.js` before merge then?

brettz9:
Sure, done

brettz9:
Oops--sorry about that JS/JSON mixup--getting late over here... :)

package.json
```diff
@@ -7,7 +7,13 @@
   "unpkg": "./umd/index.all.min.js",
   "types": "index.d.ts",
   "author": "atomiks",
+  "contributors": [
+    "Brett Zamir"
+  ],
   "license": "MIT",
+  "bugs": "https://github.com/atomiks/tippyjs/issues",
+  "homepage": "https://atomiks.github.io/tippyjs/",
+  "engines": {},
```
KubaJastrz: Leaving this empty is not very helpful, in fact npm docs suggest to set it if the engines option is present https://docs.npmjs.com/files/package.json#engines

brettz9: Hmm, ok. With these lines:

> If you specify an “engines” field, then npm will require that “node” be somewhere on that list. If “engines” is omitted, then npm will just assume that it works on node.

I guess I was confused because it doesn't end with "assume that it works on any version of Node", so I guess I had misread that as suggesting that to indicate no Node support (as in browser only), one should include it. It seems a more clear way to indicate no Node support too, but it appears one is forced to imply it works on Node. :(

Shall I revert then?

brettz9: I've asked at https://npm.community/t/engines-and-non-node-usage/6478

brettz9: Though if it takes a while to get a response or resolution, maybe I can just revert for now (or keep as is if you prefer), and possibly add it back in later in another PR if a clarification or change is forthcoming.

PR #459, opened 2019-03-27
Rollup

### LINKED ISSUE #457: Bundle popper.js in ESM builds and/or provide version without dependency
Distribution files are generally supposed to be ready for consumption without the need for build steps, but with recent releases of tippy.js, one must perform some Rollup or the like due to a dependency on "popper.js".

Please add back at least one distribution which bundles popper.js and/or which removes it. (One might use such as https://github.com/jetiny/rollup-plugin-re to strip out optional import directives for such cases.)

And thanks for the great library!

brettz9:
Builds on previous linting PR #458 to add Rollup for Popper-bundled files, including `main` and `module` updates...  Fixes #457

### DIFF
--- a/rollup.build.js
+++ b/rollup.build.js
@@ -57,10 +57,10 @@ const createPluginSCSS = output => {
   })
 }
 
-const createRollupConfigWithoutPlugins = input => plugins => ({
+const createRollupConfigWithoutPlugins = (input, {includeExternal} = {}) => plugins => ({
   input,
   plugins,
-  external: ['popper.js'],
+  external: includeExternal ? ['popper.js'] : null,
 })
 
 const createPreparedOutputConfig = format => (file, { min = false } = {}) => {
@@ -75,9 +75,11 @@ const createPreparedOutputConfig = format => (file, { min = false } = {}) => {
 }
 
 const getRollupConfigs = {
-  css: createRollupConfigWithoutPlugins('./build/css.js'),
-  index: createRollupConfigWithoutPlugins('./build/index.js'),
-  all: createRollupConfigWithoutPlugins('./build/all.js'),
+  css: createRollupConfigWithoutPlugins('./build/css.js', {includeExternal: true}),
+  index: createRollupConfigWithoutPlugins('./build/index.js', {includeExternal: true}),
+  all: createRollupConfigWithoutPlugins('./build/all.js', {includeExternal: true}),
+  indexWithPopper: createRollupConfigWithoutPlugins('./build/index.js'),
+  allWithPopper: createRollupConfigWithoutPlugins('./build/all.js'),
 }
 
 const getOutputConfigs = {
@@ -102,9 +104,13 @@ const build = async () => {
 
   const bundles = {
     index: await rollup(getRollupConfigs.index(pluginConfigs.index)),
+    indexWithPopper: await rollup(getRollupConfigs.indexWithPopper(pluginConfigs.index)),
     indexMin: await rollup(getRollupConfigs.index(pluginConfigs.indexMinify)),
+    indexWithPopperMin: await rollup(getRollupConfigs.indexWithPopper(pluginConfigs.indexMinify)),
     all: await rollup(getRollupConfigs.all(pluginConfigs.all)),
+    allWithPopper: await rollup(getRollupConfigs.allWithPopper(pluginConfigs.all)),
     allMin: await rollup(getRollupConfigs.all(pluginConfigs.allMinify)),
+    allWithPopperMin: await rollup(getRollupConfigs.allWithPopper(pluginConfigs.allMinify)),
   }
 
   // Standard UMD + ESM
@@ -120,6 +126,21 @@ const build = async () => {
     bundles.indexMin.write(outputConfigs.indexMin)
     bundles.all.write(outputConfigs.all)
     bundles.allMin.write(outputConfigs.allMin)
+
+    if (outputConfigs.index.format !== 'esm') {
+        continue;
+    }
+
+    const withPopperOutputConfigs = {
+      indexWithPopper: getOutputConfig('index.popper.js'),
+      indexWithPopperMin: getOutputConfig('index.popper.min.js', { min: true }),
+      allWithPopper: getOutputConfig('index.popper.all.js'),
+      allWithPopperMin: getOutputConfig('index.popper.all.min.js', { min: true }),
+    };
+    bundles.indexWithPopper.write(withPopperOutputConfigs.indexWithPopper);
+    bundles.indexWithPopperMin.write(withPopperOutputConfigs.indexWithPopperMin)
+    bundles.allWithPopper.write(withPopperOutputConfigs.allWithPopper)
+    bundles.allWithPopperMin.write(withPopperOutputConfigs.allWithPopperMin)
   }
 
   console.log(green('Bundles complete'))
--- a/rollup.build.js
+++ b/rollup.build.js
@@ -102,16 +102,19 @@ const build = async () => {
 
   console.log('CSS done')
 
-  const bundles = {
-    index: await rollup(getRollupConfigs.index(pluginConfigs.index)),
-    indexWithPopper: await rollup(getRollupConfigs.indexWithPopper(pluginConfigs.index)),
-    indexMin: await rollup(getRollupConfigs.index(pluginConfigs.indexMinify)),
-    indexWithPopperMin: await rollup(getRollupConfigs.indexWithPopper(pluginConfigs.indexMinify)),
-    all: await rollup(getRollupConfigs.all(pluginConfigs.all)),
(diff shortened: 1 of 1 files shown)


brettz9:
I've rebased on the new linter PR.

I've also undone a commit I had added for tying `main` and `module` to the new builds as bundling consumers might not want to tied to the particular version of popper.js and no doubt would wish to strip out redundant copies of popper.js if they had any. But they can still take advantage of the bundled builds from this PR by referencing the files directly as paths.

atomiks:
Shouldn't the files changed in the linter PR be removed here?

My only problem here is all the different combinations (or permutations?) of files, esp. including sourcemaps, bloats the npm install size. I know not many people think or care about that much though 😅, but lots of peple complain about the size of node_modules

How big is it before/after?

brettz9:
If you can review the linter PR first, that'd be easier on me--and if it needs changes, I can rebase this PR. But you can see what the incremental change was by just looking at the commit, i.e., 6ede271d9e83a19513784f095e1973f4117a0af5

Here are the total folder size differences:

esm before: 553 KB
esm now: 1.9 MB

umd before: 586 KB
umd now: 1.9 MB

brettz9:
I've rebased now off the merged linting PR.

brettz9:
Was that info ok on the file sizes? Anything else you need from me on this? Can you live with the file size increases?

atomiks:
Let's leave out the popper bundle for the UMD, it's only necessary for ESM correct?

brettz9:
Yeah, actually, hadn't thought of that. :) It might be a convenience for some, but they're injecting script tags into HTML anyways...  Will prepare right now...

brettz9:
Ok, I've amended to avoid bundling Popper for UMD versions. I've also added an extra commit (which I can squash if you prefer) which optimizes by putting the now larger set of Promises into a `Promise.all` for performance's sake.

brettz9:
Sorry just added a tweak, but now should be all set...

atomiks:
Looks like we need to apply Prettier to `rollup.build.js` too 🧐 

It works fine, my only problem is the build file seems to be spiralling out of control with complexity 😓(which was also my fault with the architecture before)

Maybe we can architect it a bit better?

brettz9:
Sure, I thought of refactoring, but didn't want to do a bunch of work if you were particular about that way.

Also, I'm not sure it would reduce "complexity" as much as it would reduce redundancy if that is what you are going for... But I can try to put something together to reduce redundancy if that's what you want...

brettz9:
I've pushed a new commit to reduce duplication considerably... Let me know if something like that is what you had in mind...

atomiks:
Yeah tbh I think that increased complexity haha

I think what you had before is good enough! I'll update and merge it

brettz9:
Excellent! Thank you... And, would it be possible for you to provide a new npm release with this?

atomiks:
Was kinda waiting for latest popper.js to release since we'll be bundling it 📦 

(That's the main problem I had with bundling btw: leads to outdated dependencies + can't use Popper.js by itself `:\`)

Something like `@pika/web` is meant to solve this problem of referencing external dependencies

brettz9:
I see, ok, gotcha.

Re: `@pika/web`, that looks interesting, but from what it seems to me, that tool would only let you release your own updates that bundle popper.js faster (or possibly relate to a latest package on unpkg?). I don't get how it supposedly gets the latest version bundled. If you have to use the command line, there is still need for work by the maintainer to update it, and the only way I can imagine the web can auto-update is if you point to an external site, whether a site like possibly unpkg, or the maintainer's website (though the latter could suffer from performance, and the former might bring trust issues for introducing another middle-man beyond npm and its dependencies). I should study it more.

atomiks:
Well it just uses the npm package files, it'll use the latest versions as specified in package.json. Since popper isn't bundled, it gets installed separately and you always get the latest version without tippy needing to update all the time.

brettz9:
Gotcha, I think--so it does the bundling for the user (with those latest allowed npm deps), so there is not so much headache for them...

atomiks:
@brettz9 

Considering reverting this PR because bundling Popper.js is really not something I'd like to do. 

It has two major limitations that make the files not worth using at all (why not use the UMD versions in the browser, which create the `tippy` and `Popper` globals that can be used in module files anyway)

- Can't use Popper constructor without downloading Popper.js twice (since Tippy will use the bundled version)
- Popper.js can be outdated

There are already two solutions for ES modules inside the browser:

- `unpkg.com?module` (no build step required, but not minified yet)
- `@pika/web` (custom simple build step I guess)
- and possibly a future spec that fixes this external dependency problem.

brettz9:
I believe there are two basic issues here, and they are quite often at odds, but I hope both can be accommodated in a manner by which those preferring to lean toward one side can do so.

One is efficiency. For this, one needs a bundler and a single pipeline for it. This avoids duplication. It is comparable to the likes of C++ projects which require building from source to get the latest benefits of a working version which, though stable, has not yet been prepared as pre-compiled binaries for the convenience of other of their users.

The other is genuine modularity and lack of need for concern for implementation details (like the pre-built binaries in the above example), a particular challenge for the web where there is no official central repository for the ecosystem (unlike Node) and modules come from remote sources.

For this, the essential element is that the includer of a module should not need to be concerned with, or even know about, where its sources come from. The module should be a self-sufficient black box (with the exception of modules which set globals for polyfill purposes, and even here, imo, polyfills of globals, as with `@babel/polyfilll` is one case where it logically makes sense to put a script tag in HTML, as it is really a dependency of a backward browser, rather than one of any particular module).

The only requirement of consumers should be to import the module (from their own module(s), which may be nested several levels away from the main HTML script) and use its exposed methods/properties.

If I have to include dependency scripts for it in my HTML, besides the added trouble for me to get started with your code, I am polluting my HTML with aspects beyond my HTML's direct concern and subject to having to update my HTML again if one of the dependencies or their paths change in the future, whereas, if I can just "set and forget" in pointing at a self-contained module of interest, I will happily get the benefit of updates down the road. If I am ever really in need of having the absolute latest dependencies (like the person not content with a somewhat outdated binary), I can switch to doing my own bundling and non-modularly force myself into concern with the details of your implementation, such as that I must specify Popper in my chain, and how to bundle it, etc.

Unpkg is not a standard feature of the web, requiring trust of the site after build time, and @pika/web still requires building however simple (and not familiar or used by all browser users).

As far as the non-availability of the `Popper` constructor, you could add it as a named export, so the version which is bundled can also be (optionally) utilized.

atomiks:
Thanks for the reply

> If I have to include dependency scripts for it in my HTML, besides the added trouble for me to get started with your code, I am polluting my HTML with aspects beyond my HTML's direct concern and subject to having to update my HTML again if one of the dependencies or their paths change in the future

Tippy will only ever have a single dependency for UMD builds, so there will only be two script tags to use: https://atomiks.github.io/tippyjs/getting-started/#option-1-cdn. If there were 3+ that would be a concern, especially if they didn't matter (e.g. polyfills or something).

The benefit is that you probably would like to use Popper.js for other parts of your app (e.g. menu systems which Tippy isn't designed to do), so maybe thinking about its existence is a good thing. Of course, when using it in traditional build systems, it already is a black box since you only `import tippy from 'tippy.js'` and install one dependency.

Additionally, using `@major` with unpkg allows auto-updates to the latest non-breaking version without needing to think about it. 

> As far as the non-availability of the Popper constructor, you could add it as a named export, so the version which is bundled can also be (optionally) utilized.

This is a great point but the remaining concern is still about it being outdated unfortunately...

brettz9:
Re: extra script tags, I am just particular in hoping my HTML has *no* script tags at all besides my own application and any polyfills for older browsers, and also a version which works locally without need for a build step for rapid debugging. One path to an import script make sense, if it is in the file dependent on them.

But it's not the end of the world. I've lived with it so far as is. :)

And though it would have been nicer to be spared all of the work, I know one cannot always think of all of the possible concerns ahead of time, and I hope at least the linting was of use. Best wishes...

atomiks:
Yeah, I wish there were a better solution here... your work was appreciated regardless so thank you.

brettz9:
Glad to hear it--and thanks for the awesome library!

brettz9:
Btw, FWIW, I realized that I could at least do my Rollup in a far more simplified and theoretically less fragile manner, by rolling up your distributed ESM file instead of bothering to try to copy or replicate the complicated build file (i.e., it wasn't a binary choice between compiling from your source or using script tags). 

I only needed to add the node-resolve plugin, targeting your ESM dist file as my input, so I can create a local bundled ESM copy. Besides allowing me to work with live copies of my own app without needing a build step or with polluting my HTML (or needing "magic" in my JS, reaching for globals), I do get the benefit admittedly of being able to get the latest popper.js now whenever I wish as well--though without needing to care about any other dependencies or routines you may add in source that end up pre-bundled into the distributed copy. And if I need the Popper.js constructor, I can worry about transforming my routine to get that too.

Issue #460, opened 2019-04-04
A virtual positioning referance is not working with delay


kGk13:
Hello,

I've used your example from codepen for a vistual positioning referance, all the options are working except the delay. I'm doing something wrong ? This is the code that i'm using:

`
function initTippy(els){
    for (let index = 0; index < els.length; ++index) {
        demo(els[index]);
    }
}
initTippy(document.querySelectorAll('article #post-content .text a, .birdviewminimal-container .categories a'));

function demo(anchor){
    anchor.addEventListener('mouseenter', show);
    anchor.addEventListener('mouseleave', hide);
    anchor.addEventListener('focus', event => {
        if (!instance.state.isVisible) {
        show(event)
    }
    })
    anchor.addEventListener('blur', hide)

    function createVirtualReference() {
        return {
            getBoundingClientRect: () => anchor.getBoundingClientRect(),
            clientWidth: anchor.clientWidth,
            clientHeight: anchor.clientHeight,
    }
    }

    const virtualReference = createVirtualReference()

    const template = document.getElementById('tooltip-template');
    const container = document.createElement('div');
    container.appendChild(document.importNode(template.content, true));

    function show(event) {

        if (event.type === 'mouseenter') {

            const LINE_HEIGHT = 24
            const rect = anchor.getBoundingClientRect()
            const cursorPoint = Math.round(event.clientY - rect.top)
            const lineIndex = Math.floor(cursorPoint / LINE_HEIGHT)

            const top = rect.top + lineIndex * LINE_HEIGHT
            const bottom = top + LINE_HEIGHT

            Object.assign(virtualReference, {
                clientHeight: bottom - top,
                delay: [1600, 0],
                getBoundingClientRect() {
                    return {
                        width: 0,
                        height: bottom - top,
                        top,
                        bottom,
                        //delay: 10000,
                        delay: [1600, 0],
                        left: event.clientX,
                        right: event.clientX,
                    }
                },
            })

            // Prevent the tooltip from following scroll
            setTimeout(() => {
                instance.popperInstance.disableEventListeners()
        })
        } else {
            // For `focus`, we don't have client coords. Just use the anchor.
            Object.assign(virtualReference, createVirtualReference())
        }

        instance.show()

    }

    function hide() {
        instance.hide()
    }

    const instance = tippy(virtualReference, {
        content: container.innerHTML,
        arrow: true,
        theme: 'light',
        arrowType: 'round',
        placement: 'top',
        boundary: 'window',
[...]

PR #461, opened 2019-04-05
Expose schedule{Hide,Show} methods

atomiks:
As surfaced by #460, `delay` / `wait` don't work for virtual references when calling `.show()` / `.hide()` since they are immediate. Like the internal event handlers, users should be calling the schedule methods instead for virtual references

TODO:

- Should they accept a `duration` argument like `show/hide`?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -101,6 +101,8 @@ export interface Instance {
   popperInstance: PopperInstance | null
   props: Props
   reference: ReferenceElement
+  scheduleHide(): void
+  scheduleShow(event?: Event): void
   set(options: Options): void
   setContent(content: Content): void
   show(duration?: number): void
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -114,6 +114,8 @@ export default function createTippy(
     setContent,
     show,
     hide,
+    scheduleShow,
+    scheduleHide,
     enable,
     disable,
     destroy,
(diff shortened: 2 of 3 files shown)


KubaJastrz:
Isn't it a little confusing that some methods respect `delay`/`wait` and some don't? I think the better solution is to add a flag parameter to `show`/`hide` if you want to prevent making breaking changes, instead of exposing internal methods.

atomiks:
It would be weird to call `.show()` and have it respect delays since a programmatic call is more like synchronous behavior. And since delays are only really relevant for UI events.

Do you mean something like:

`.show(true)`
vs
`.scheduleShow()`

KubaJastrz:
Yes. I guess it makes sense if you put it that way. Shouldn't there be a callback mechanism for scheduleShow then?

atomiks:
What would the callback do?

KubaJastrz:
Call stuff after scheduled show method actually runs. But I guess you can reuse `onShow` prop.

atomiks:
Your comment there made me think of the `update()` callback workaround I need to do for Popper. Decided to make it better.

Still, not sure if Tippy needs it, but I can see Popper needing it

atomiks:
Since:

- These methods will mainly be used in event handlers for virtual references
- `onShow()` exists

I don't see a need for `duration` and the callback. Going to merge...

PR #467, opened 2019-04-16
Fix multiple tippys if using touch

atomiks:
When using a touch device with `hideOnClick: true`, only a single tippy displays when multiple tippys are attached to the reference element

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -134,7 +134,7 @@ export interface PopperChildren {
 export interface HideAllOptions {
   checkHideOnClick?: boolean
   duration?: number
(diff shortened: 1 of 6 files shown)


index.d.ts
```diff
@@ -134,7 +134,7 @@ export interface PopperChildren {
 export interface HideAllOptions {
   checkHideOnClick?: boolean
   duration?: number
-  exclude?: Instance
+  exclude?: any
 }
```
atomiks: I don't know how to type this to work with the new `hideAll` behavior, cc @KubaJastrz

PR #470, opened 2019-04-17
followCursor fixes

### LINKED ISSUE #469: Interactive and followCursor allow tooltip to be moved from within tooltip
If you set both `interactive: true` and `followCursor: 'horizontal'`, when your mouse enters into the tooltip, you can still move the tooltip: https://codepen.io/smargh/pen/vMpzMK

It would make more sense to me if `followCursor` only applied when the cursor was moving within the tooltip parent, not the tooltip itself.

atomiks:
- `followCursor` currently does not respect `popperOptions.modifiers.preventOverflow.padding` and instead only uses the hardcoded `PADDING` constant, this change makes it respect it
- Resolves #469, prevents the tooltip from moving while the cursor is on top of it. This changes the behavior slightly if the tippy is interactive, since it won't continue to follow the cursor while it's still hiding (unlike it currently does)

Since this change adds some size, found a way to also cut back on the `closest` polyfill by reusing the `closestCallback` function

cc @fractaledmind - I don't think this behavior is exactly what you suggested (to only follow if it's over the parent/popper node) as it only follows if it's over the reference element now.

### DIFF
--- a/src/bindGlobalEventListeners.ts
+++ b/src/bindGlobalEventListeners.ts
@@ -54,7 +54,7 @@ export function onDocumentClick(event: MouseEvent): void {
   }
 
   // Clicked on a reference
-  const reference: ReferenceElement | undefined = closestCallback(
+  const reference: ReferenceElement | null = closestCallback(
     event.target,
     (el: ReferenceElement) => el._tippy && el._tippy.reference === el,
   )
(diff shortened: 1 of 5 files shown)


fractaledmind:
I actually think I was being loose with my language; having the tooltip only follow the mouse when the mouse is over the reference element indeed makes the most sense.

atomiks:
Alright yeah, this seems to work well then.

Issue #482, opened 2019-05-01
Ability to pass custom style tag


jasondibenedetto:
I'm dealing with a specific case where I need to be able to query the style tag generated by tippy. Does anyone see any problems with adding an option here to allow for a custom style tag to be passed through, so that a class or id attribute could be added.

https://github.com/atomiks/tippyjs/blob/dce36f820ce22d0a68a21f26fc387b0706c155a7/src/css.ts#L6-L19

atomiks:
Interested to hear your specific case?

Also related: #480 - seems like this function could use some improvement

jasondibenedetto:
I'm currently working on app that requires calling [stylesheet.disabled](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet/disabled) to remove some unwanted stylesheets (It's not ideal, but it's just the way it needs to be unfortunately). 

I'm using https://github.com/atomiks/tippy.js-react, and I just need a bit more control over the stylesheet element so I can prevent it from being disabled with the rest of the unwanted stylesheets. e.g.

```
<style type="text/css" class="app">.tippy { ... }</style>
```

I know it's a bit of an edge case feature, so I thought I'd just ask first.

atomiks:
Can a hardcoded selector work:

`class="tippy-stylesheet"` (className)
or 
`data-tippy-stylesheet` (attribute)

Not sure which would be better?

jasondibenedetto:
Yeah that should work as well. I would be happy either way, maybe class is more suited?

jasondibenedetto:
Would you like me to make a PR?

atomiks:
Ha yes was going to say if you wanted to, I think it's only one line but you can be added to the contributors 🙂

jasondibenedetto:
https://github.com/atomiks/tippyjs/pull/483

PR #483, opened 2019-05-01
Add data attribute to style tag

jasondibenedetto:
See https://github.com/atomiks/tippyjs/issues/482

Decided on using a data attribute instead, after seeing styled-components using `<style data-styled />` in their build.

### DIFF
--- a/src/css.ts
+++ b/src/css.ts
@@ -8,6 +8,7 @@ export function injectCSS(css: string): void {
     const style = document.createElement('style')
     style.type = 'text/css'
     style.textContent = css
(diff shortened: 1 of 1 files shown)


atomiks:
Thanks!

lk77:
Hello,

there is a way to remove that style tag ?

we do not want css styling inside the head, because it become a mess quickly, we prefer to link to a css file.

atomiks:
@lk77 yep, [info available here](https://atomiks.github.io/tippyjs/getting-started/#css)

Issue #484, opened 2019-05-01
issues with including tippy.js as a 3rd party library


mreinstein:
Tippy is amazing! I've used it on a few personal websites. 

I'd like to be able to use it in a 3rd party library ( a widget that is embedded on someone else's website.) One of the requirements of doing this is being able to namespace css declarations to reduce the likelihood of polluting global scope. Some questions around this:

* would it be possible to namespace all of the classes tippy declares? e.g., `.CUSTOM_NAMESPACE-tippy-tooltip { ... }`
*  Is it possible to embed another style into the tippy bundle (I don't want the dark theme baked in, I want one of the light themes.)

much thanks for a phenomenally good component!

atomiks:
`tippy` acts as its own namespace (apart from `.dark-theme`, in that case you can set `theme: ''` or something else. For the light theme you need to reference its stylesheet separately

mreinstein:
doesn't tippy already include the dark theme by default? It gets embedded into the tippy bundle, right?
Is there a way for me to instead build tippy with another theme embedded and set to the default?

atomiks:
There's no `dark-theme` selector, that's just the base styling of `.tippy-tooltip`. The CSS is only 1.6 kB minzipped, and there are no selectors in the CSS not prefixed with `.tippy-*` so conflicts should be minimal

mreinstein:
I hate to be a pest, but I think maybe the use case isn't clear. 

I'm embedding my component in someone else's website. That website has lots of other content on it that I don't control:
* if the the website already has tippy components on it, and they are themed differently, I don't want to affect the look of those, because it's someone else's content.
* these websites really don't like having their 3rd party integrations inject unprefixed styles directly into their documents. My organization has an assigned prefix (`e.g., all my classes and ids must start with `acme123-`)
* a given website frequently has several or dozens of these 3rd party components. asset space is a premium. I know 1.6 kb minzipped is very light, but it adds up.

Anyway, I know this is an extremely uncommon use case, I don't expect most people will run into these problems. I'm not expecting a big engineering effort to support this use case. I was just hoping that given there's not much css that maybe I could build my own copy by passing some parameters to rollup to namespace stuff, or at least replace the embedded theme.

Thanks again!

atomiks:
The problem is all the elements have the `tippy-*` prefix hardcoded in the JS and CSS. If you don't want to run into edge cases then unfortunately a fork might be needed. 

[...]

PR #487, opened 2019-05-03
Use namespace prefix for custom builds

### LINKED ISSUE #486: hardcoded tippy css class references in popper.ts
I noticed in https://github.com/atomiks/tippyjs/blob/master/src/popper.ts we hardcode a few strings for specific tippy classes (`tippy-roundarrow`, `tippy-arrow`, `tippy-backdrop`, etc.)

Does it make sense to move these strings to `constants.ts` ?

atomiks:
Closes #486 @mreinstein

Have to use `'tippy'` as default otherwise tests break

### DIFF
--- a/rollup.build.js
+++ b/rollup.build.js
@@ -12,6 +12,7 @@ const resolve = require('rollup-plugin-node-resolve')
 // const commonjs = require('rollup-plugin-commonjs')
 const json = require('rollup-plugin-json')
 const cssOnly = require('rollup-plugin-css-only')
+const replace = require('rollup-plugin-replace')
 const { green, blue } = require('colorette')
 
 const BANNER = `/**!
@@ -20,12 +21,18 @@ const BANNER = `/**!
 * MIT License
 */`
 
+const NAMESPACE_PREFIX = process.env.NAMESPACE || 'tippy'
+
 const extensions = ['.js', '.ts']
 
 const plugins = {
   babel: babel({
     exclude: 'node_modules/**',
     extensions,
+  }),
+  replace: replace({
+    "NAMESPACE_PREFIX = 'tippy'": `NAMESPACE_PREFIX = '${NAMESPACE_PREFIX}'`,
+    delimiters: ['', ''],
   }),
   minify: terser(),
   resolve: resolve({ extensions }),
@@ -38,7 +45,7 @@ const BASE_OUTPUT_CONFIG = {
   globals: { 'popper.js': 'Popper' },
   sourcemap: true,
 }
-const BASE_PLUGINS = [plugins.resolve, plugins.json]
+const BASE_PLUGINS = [plugins.replace, plugins.resolve, plugins.json]
 
 const pluginConfigs = {
   index: [plugins.babel, ...BASE_PLUGINS],
@@ -50,6 +57,9 @@ const pluginConfigs = {
 const createPluginSCSS = output => {
   return sass({
     output,
+    options: {
+      data: `$namespace-prefix: ${NAMESPACE_PREFIX};`,
+    },
     processor: css =>
       postcss([autoprefixer, cssnano])
         .process(css, { from: undefined })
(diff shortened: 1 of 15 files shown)


mreinstein:
@atomiks I think you mean #484 ?

rollup.build.js
```diff
@@ -20,13 +21,19 @@ const BANNER = `/**!
...
+
 const extensions = ['.js', '.ts']
 
 const plugins = {
   babel: babel({
     exclude: 'node_modules/**',
     extensions,
   }),
+  replace: replace({
+    "NAMESPACE_PREFIX = 'tippy'": `NAMESPACE_PREFIX = '${NAMESPACE_PREFIX}'`,
```
mreinstein: This could be easier to read with something like `"__NAMESPACE_PREFIX__": NAMESPACE_PREFIX,`

Then in the code it could be referenced like this:

```javascript
export const IOS_CLASS = '__NAMESPACE_PREFIX__-iOS'
```
The big benefits here is
*  it looks very obvious when viewing the source that `__NAMESPACE_PREFIX__ is a value that gets replaced at build time. As opposed to `NAMESPACE_PREFIX` which just looks like a standard javascript variable.
* there's less variables in the built code that ships to the browser. If someone were to look at the built version of tippy, they'd see `export const IOS_CLASS = 'tippy-iOS'`

src/popper.ts
```diff
@@ -79,13 +85,13 @@ export function createArrowElement(
 ): HTMLDivElement {
   const arrow = div()
   if (arrowType === 'round') {
-    arrow.className = 'tippy-roundarrow'
+    arrow.className = selectorToClassName(ROUND_ARROW_SELECTOR)
```
mreinstein: instead of introducing a new function here, maybe we could create another constant for this?
e.g., we could have `ROUND_ARROW_CLASS` *and* `ROUND_ARROW_SELECTOR`

src/constants.ts
```diff
@@ -10,13 +13,13 @@ export const PLACEMENT_ATTRIBUTE = 'x-placement'
...
+export const ACTIVE_CLASS = `${NAMESPACE_PREFIX}-active`
 
 // Selectors
-export const POPPER_SELECTOR = '.tippy-popper'
-export const TOOLTIP_SELECTOR = '.tippy-tooltip'
-export const CONTENT_SELECTOR = '.tippy-content'
-export const BACKDROP_SELECTOR = '.tippy-backdrop'
-export const ARROW_SELECTOR = '.tippy-arrow'
-export const ROUND_ARROW_SELECTOR = '.tippy-roundarrow'
+export const POPPER_SELECTOR = `.${NAMESPACE_PREFIX}-popper`
```
mreinstein: how about:

```javascript
export const POPPER_CLASS = '__NAMESPACE_PREFIX__-iOS'
export const POPPER_SELECTOR = `.${POPPER_CLASS}`
```

mreinstein:
@atomiks  PR looks dope! I added some comments on it. All of these focus on the goal of making the namespacing have 0 impact on the built tippy files. (the namespace stuff would be baked in at build time and incur no function calls or additional string interpolation.)

much thanks!

src/popper.ts
atomiks: Was trying to save as many bytes as possible

rollup.build.js
atomiks: Yeah I avoided it because of tests but I've updated them now 🙂

atomiks:
Great suggestions @mreinstein! I just had to update tests without hardcoding to use your method. This works great now

atomiks:
Should we also apply this prefix to #483 (`data-tippy-stylesheet`)?

mreinstein:
Code looks super clean! I'm glad we found a way to do it that has 0 run-time impact on tippy. :) Looking forward to replacing the janky popover code I'm using with this!

> Should we also apply this prefix to #483 (data-tippy-stylesheet)?

Yeah I think that makes sense, I didn't catch this the first time.

PR #511, opened 2019-06-03
Backport v5 fixes to v4

### LINKED ISSUE #504: consider a more specific selector for pointer-events
This particular styling: https://github.com/atomiks/tippyjs/blob/master/src/scss/index.scss#L144-L146

What's happening for me is I have an tippy with html, interactive content. That content contains an SVG file. My svg file's `<path>` children are getting set to `pointer-events: auto;`.

Would it be easy to use a more specific selector so that this style won't be applied to my content?

### LINKED ISSUE #509: Scroll bar flicker in IE11 when a container's height is 100%
In the case when a container's height is 100%, when you hover over the tippy reference in IE11, a scroll bar appears then immediately disappears. Is there any way to solve this problem?

**How to reproduce:**
1. Open https://jsfiddle.net/12amsko5 in IE11
2. Hover over some box

**Preview:**
Preview
(low fps, in real case reproduced more often)

atomiks:
The first commit has the async behavior change when mounting, which if they were testing, could break their tests. But I think this is an implementation detail of tippy, no?

(edited: it does work correctly on IE11)

### Fixes

- Mounting behavior (IE11 scrollbar flicker) (fix #509)
- `followCursor` fixes (respects `boundary` & fix regression where `initial` was not placed correctly on touch devices on first show)
- Ensure `destroy()`'s unmounting of the tippy can never be impeded
- Fix the longstanding interactive scrolling issue
- Fix `path` selector having `pointer-events: auto` (fix #504)

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -72,7 +72,6 @@ export default function createTippy(
   let hideTimeoutId: number
   let animationFrameId: number
   let isScheduledToShow = false
-  let currentParentNode: Element
   let previousPlacement: string
   let wasVisibleDuringPreviousUpdate = false
   let hasMountCallbackRun = false
@@ -226,13 +225,14 @@ export default function createTippy(
   }
 
   /**
-   * Determines if the instance is in `followCursor` mode
+   * Determines if the instance is in `followCursor` mode.
+   * NOTE: in v5, touch devices will use `initial` behavior no matter the value.
    */
-  function hasFollowCursorBehavior(): boolean {
+  function getIsInLooseFollowCursorMode(): boolean {
+    const { followCursor } = instance.props
     return (
-      instance.props.followCursor &&
-      !isUsingTouch &&
-      lastTriggerEventType !== 'focus'
+      (followCursor && lastTriggerEventType !== 'focus') ||
+      (isUsingTouch && followCursor === 'initial')
     )
   }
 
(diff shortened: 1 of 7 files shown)


atomiks:
> I'm not 100% sure about the requestAnimationFrame support in tests but it looks like you got it working with jest.runAllTimers(), so if anyone needs it, they can use that too anyway.

Test setup file is:

```js
// We want to use macrotask timers that can be mocked by Jest
global.Promise = require('promise')

global.nativeRequestAnimationFrame = requestAnimationFrame
global.requestAnimationFrame = cb => cb()
```

atomiks:
Looks like there's an unmounting bug if moving the cursor quickly over elements (probably in v5 branch as well.) This new async behavior needs some more testing before I merge this 🤒

atomiks:
Turns out Firefox scrollbar appearing thing didn't get fixed. Don't really know what's up there and never have since v0 🤷‍♂ 

IE11 is still fixed (unless the reference element is right next to/under the scrollbar as mentioned in the issue. Don't know if that edge case can be resolved, but this is at least better in vast majority of cases).

Looks like we're cancelling any async scheduling of frames and timeouts so I _hope_ there's no problems or regressions there. I'll probably postpone v5 until the end of the month in case there are any regressions with this patch release.

PR #514, opened 2019-06-03
Fix #513

### LINKED ISSUE #513: *-start/*-end placement with followCursor option reversed + incorrect offset (v4.3.2)
## Bug description

Thanks for the backport v5 fixes to v4 (#511).
In the case where the `placement` option has `*-start`/`*-end` value and `followCursor` option is defined, tippy is placed on the other side of the cursor in 4.3.2 version. Is this a new expected behavior or bug?

## Reproduction

CodePen link (4.3.1): https://codepen.io/lozinsky/pen/MdLzwo
CodePen link (4.3.2): https://codepen.io/lozinsky/pen/JqxeYa

atomiks:
This fixes the problem where the tippy gets offset incorrectly from the cursor position

However, the reversal of the placement variation is now actually correct with the new behavior... but it's unexpected when considering it's a patch release. Unfortunately I don't think that can be changed back now.

- The size of the virtual reference should match the popper element's size
- The variation needs to be taken into consideration

@lozinsky can you verify

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -407,14 +407,18 @@ export default function createTippy(
       getBasicPlacement(popper),
     )
     const fullPlacement = popper.getAttribute(PLACEMENT_ATTRIBUTE)
-    const isVariation = !!(fullPlacement ? fullPlacement.split('-')[1] : false)
+    const isVariation = fullPlacement ? !!fullPlacement.split('-')[1] : false
     const size = isVerticalPlacement ? popper.offsetWidth : popper.offsetHeight
     const halfSize = size / 2
     const verticalIncrease = isVerticalPlacement
       ? 0
-      : halfSize + (isVariation ? halfSize : 0)
+      : isVariation
+      ? size
+      : halfSize
     const horizontalIncrease = isVerticalPlacement
-      ? halfSize + (isVariation ? halfSize : 0)
+      ? isVariation
+        ? size
+        : halfSize
       : 0
 
     if (isCursorOverReference || !instance.props.interactive) {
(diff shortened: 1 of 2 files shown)


src/createTippy.ts
```diff
@@ -402,14 +402,20 @@ export default function createTippy(
...
     // The virtual reference needs some size to prevent itself from overflowing
-    const fakeSize = 100
-    const halfFakeSize = fakeSize / 2
     const isVerticalPlacement = includes(
       ['top', 'bottom'],
       getBasicPlacement(popper),
     )
-    const verticalIncrease = isVerticalPlacement ? 0 : halfFakeSize
-    const horizontalIncrease = isVerticalPlacement ? halfFakeSize : 0
+    const fullPlacement = popper.getAttribute('x-placement')
```
lozinsky:
```suggestion
    const fullPlacement = popper.getAttribute(PLACEMENT_ATTRIBUTE)
```

test/spec/props.test.js
```diff
@@ -1145,8 +1145,8 @@ describe('followCursor', () => {
   // should be within that
   const first = { clientX: 317, clientY: 119 }
   const second = { clientX: 240, clientY: 500 }
-  const fakeSize = 100
-  const halfFakeSize = fakeSize / 2
+  const size = 0
+  const halfSize = size / 2
```
lozinsky: 0 / 2? It is for documentation?

atomiks: `.offsetWidth/Height` are 0 in Jest since it doesn't render anything... just copying the source code

PR #499, opened 2019-05-20
[WIP] V5

atomiks:
# Tippy v5

The main goal of this version is to:

1. Improve developer experience with warnings without bloating production bundle size
2. Massively reduce core size and make the library more treeshake-able
3. Allow new feature additions to be added separately without increasing bundle size
4. Improve naming consistency and usage

## Highlights

- ⬇️ 30%+ smaller download size (5.4 kB minzipped including core CSS)
- ⬇️ 50%+ smaller parse size (15.1 kB minified including core CSS)
- ⬇️ 60%+ smaller core CSS (0.6 kB minzipped) + smaller external animation files
- ⚙️ Development and production versions for better developer experience
- ✨ New animation variations
- ✨ New `touch: ["hold", delay]` prop (for long press behavior)
- ✨ New `arrow: string | Element` values to use your own shape
- ✨ New `createSingleton` method, supersedes `group()`
- 🌸 Improved animations integration and documentation, namely fully dynamic transition dimensions when tippy content updates
- 🌸 Improved naming consistency
- ♿ Improved accessibility defaults for interactive tippys
- 🐛 Various minor behavior bug fixes

## Development and production versions of Tippy.js

We're now adding helpful development warnings without bloating bundle size. Development files are `.js`, while production versions are `.min.js`.

Main filename was renamed from `index.all` to `tippy.bundle` for better DevTools display.

```
tippy.js/
├── dist/
│   ├── tippy.cjs.js              "main" file
│   ├── tippy.esm.js              "module" file
│   ├── tippy-bundle.iife.js              
│   ├── tippy-bundle.iife.min.js         "unpkg" file
│   ├── tippy.iife.js              
│   ├── tippy.iife.min.js
```

```js
// This whole block will get stripped by minifiers/DCE for production builds
// When importing like `import tippy from 'tippy.js'`
if (process.env.NODE_ENV !== 'production') {
  if (somethingBad) {
    console.warn('You did somethingBad');
  }
}
```

Where's the `umd` folder? The browser format is now `iife`. Node is covered by `cjs`, and `amd` is dead and/or didn't work with `popper.js`/`tippy.js` anyway before because of the `.js` suffix apparently (which is what `umd` covered). Now it's not necessary.

`iife` development version includes the block without the Node-only `process` variable. In the `.min` version, it's stripped out entirely (like the other formats).

---

## Drop Android 4.4 support

Remove all `-webkit-` prefixed transforms to save 200 minzipped bytes or so.

---

[...]

PR #523, opened 2019-06-22
[WIP] Add `inlinePositioning` as extra-prop

atomiks:
See https://github.com/atomiks/tippyjs/pull/499#issuecomment-504630197 for details.

Run `npm run dev:extra-props` demo for behavior.

Need to add tests for this.


atomiks:
Tests aren't complete here but need to use some new v5 work for this and will continue

Issue #543, opened 2019-07-19
sticky: true repaints frequently


mreinstein:
## Bug description
setting `sticky: true` and `trigger: click` on an element, it seems to repaint the document rather frequently, so much so that the dom can't be edited from dev console.

(This is happening in chrome, but might be happening in other browsers too.)

## Reproduction

1. visit https://test-tippy-obddnsfyyn.now.sh/

2. click the button

3. inspect the tippy element

4. notice that various dom elements in the dev console are flickering and un-editable:

![ScreenRecording2019-07-18at7](https://user-images.githubusercontent.com/718067/61505014-46397200-a992-11e9-8e39-25de5707bbb0.gif)

atomiks:
That's its purpose and perf limitation, because updates need to occur on every animation frame in case the element moves for any reason and Popper needs to immediately catch up, so there's "wasted" updates when the element doesn't move. I don't think it's possible to know exactly how and when the element changes its position on the document, so use this particular prop with caution

Wyzix33:
With 3-4 sticky tooltips open CPU is too high ... I've managed to fix this problem by calling 
```instance.popperInstance.update()``` when i needed to update tooltip position, it was a little tricky with the tooltip that was inside another tooltip ... that needs to have a lot of events attached to work without sticky: true... or i can just destroy() it when parent tooltip get's updated... still looking for a good solution to that problem.

mreinstein:
> updates need to occur on every animation frame in case the element moves for any
> reason and Popper needs to immediately catch up

It makes sense that we need to poll every animation from to detect element movement, but I'm wondering if there's a way we can only read the element state, rather than actually writing anything to the dom. That might help significantly.

atomiks:
Mm yeah I guess, but what about `transform: scale()` transitions of the reference element, or layout changes: https://codepen.io/atomiks/pen/WVQpde

`sticky` updating each animation frame fixes those cleanly without needing to check anything.

mreinstein:
> sticky updating each animation frame fixes those cleanly without needing to check anything.

I get that the static update each frame is very clean. I think the issue though is that updating essentially does 2 things under the hood each frame:
* re-draws the element
* triggers a layout re-flow

For most of those frames, nothing will have changed. I'm wondering if we could use`Element.getBoundingClientRect()` each frame to detect if the element has changed. That would still trigger layout re-flow, but at least we're not thrashing the dom with unnecessary re-draws.

[...]

PR #547, opened 2019-07-31
backport position: sticky perf improvement to master #543

mreinstein:


atomiks:
Thanks. Was trying to think of how v5 might be different to v4 in this regard but I think there's no difference. I'll do further testing to ensure, anyway.

mreinstein:
thanks for merging this! 👍

PR #557, opened 2019-08-12
Fix double whitespace in `theme`

### LINKED ISSUE #556: Double whitespace in `theme`
## Bug description

Using a double whitespace to separate class names inside of the `theme` property triggers an unwanted class.

For example, `theme: 'light  popover  popover--note'` will result in a div.tippy-tooltip with these classes:

    class="tippy-tooltip light-theme -theme popover-theme popover--note-theme"

Instead of the expected:

    class="tippy-tooltip light-theme popover-theme popover--note-theme"

i.e.: the extra space results in the creating of a `-theme` class.

Tested with tippy 4.3.4.

Not a big deal, but made things a bit awkward to read.

## Reproduction

CodePen link: https://codepen.io/fabswt/pen/rXENjE

KubaJastrz:
Fixes #556

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -45,7 +45,7 @@ import {
   setTransitionDuration,
   setVisibilityState,
   isRealElement,
-  arrayFromString,
+  splitBySpaces,
 } from './utils'
 
 let idCounter = 1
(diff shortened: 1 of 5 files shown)


KubaJastrz:
Good point, I had to leave `.filter(Boolean)` to filter out empty strings.

atomiks:
`str.trim().split(/\s+/)` leaves no empty strings tho?

KubaJastrz:
Difference in performance seems pretty negligible: https://jsperf.com/filter-empty-string/1

PR #576, opened 2019-09-23
Handle case of weird head tag layout

finom:
The PR handles cases of weird layout of <head> tag when styling tags (link, style) aren't children of <head> tag, but they're children of other children of <head> tag. The fix makes firstStyleOrLinkTag variable to ignore such weird cases.

That's the exact case I'm going to handle:

Unfortunately I don't have access to change this code and tippy initialization throws `Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.`


atomiks:
Thank you

PR #583, opened 2019-09-29
Ignore __DEV__ else branches for coverage

atomiks:
Maybe there is a plugin for this or something? I'd like to not have to write it above every `__DEV__` branch, but it's lowering branch % coverage since the "else" block doesn't matter to check here.


KubaJastrz:
I don't think there is a built-in solution. The only thing to automate this would be babel plugin that runs before tests I think.

atomiks:
I asked on Stack Overflow and they mentioned you should test "else" if you have side effects within it (although I don't think we do anywhere so it doesn't matter - maybe we might in the future)

atomiks:
Travis seems to be working again I think 👁

KubaJastrz:
I saw that SO response and after some thought I agree. This is still technically part of the library code and we can add a spy to `console.warn` to check if it is not called in production environment. Better safe than sorry I guess 🙄

PR #591, opened 2019-10-02
Add support for plugins in addons

### LINKED ISSUE #588: Addons should support plugins
## Bug description

`createSingleton()` & `delegate()` can't take plugins currently

atomiks:
Allows ability to pass plugins as 3rd arg to addons, like `tippy()`

Fixes #588 

/cc @KubaJastrz for the types


atomiks:
Verified these types in test TS project.

Adding tests for types would be very helpful for this

PR #594, opened 2019-10-04
Update MIGRATION_GUIDE.md

jasonshillingford:
Fix spelling of `tippy-bundle-iife.js` to `tippy-bundle.life.js`


jasonshillingford:
Sorry the file name is `tippy-bundle-iife.js`

atomiks:
`iife` = immediately invoked function expression

I also remember it looking weird when first looking seeing it (and thinking "misspelled **l**ife?"), but it's the name of the format like `esm` or `cjs`

JamoCA:
In 6.x, iife was replaced with umd, right?  (The v6 documentation doesn't mention this on the FAQ page.)

PR #605, opened 2019-10-15
Fix onWindowBlur activeElement broken condition

lozinsky:
Access to the properties of `activeElement` occurs before condition, which leads to an error.
For example, error in IE11: `Unable to get property '_tippy' of undefined or null reference`.

## What I did
I changed the condition, moved access to the `activeElement` properties after condition.

### DIFF
--- a/src/bindGlobalEventListeners.ts
+++ b/src/bindGlobalEventListeners.ts
@@ -1,4 +1,5 @@
 import {PASSIVE} from './constants';
+import {isReferenceElement} from './utils';
 
 export const currentInput = {isTouch: false};
 let lastMouseMoveTime = 0;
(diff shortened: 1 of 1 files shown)


src/bindGlobalEventListeners.ts
```diff
@@ -45,16 +46,14 @@ export function onDocumentMouseMove(): void {
...
-    activeElement &&
-    activeElement.blur &&
-    instance &&
-    !instance.state.isVisible
-  ) {
-    activeElement.blur();
+  if (isReferenceElement(activeElement)) {
+    const instance = activeElement._tippy!;
+
+    if (activeElement instanceof HTMLElement && !instance.state.isVisible) {
```
atomiks: `instanceof` checks don't work in iframes

atomiks: `isType(activeElement 'HTMLElement')` util should work

lozinsky: `isType` will not work, for example, for a button need to pass `'HTMLButtonElement'` as type, then `isType` returns `true`. Maybe revert `activeElement.blur` condition?

atomiks: > Maybe revert activeElement.blur condition

Seems safest option yes

PR #619, opened 2019-10-24
chore: remove `--silent` flag

KubaJastrz:
I didn't like that I couldn't use `console.log` to debug some values in tests, so I mocked `console.warn` globally.

Not sure if this is the best solution but it works 🤷‍♂

### DIFF
--- a/test/integration/createTippy.test.js
+++ b/test/integration/createTippy.test.js
@@ -179,8 +179,6 @@ describe('instance.destroy()', () => {
     jest.advanceTimersByTime(500);
 
     expect(spy).not.toHaveBeenCalled();
-
-    spy.mockRestore();
   });
 });
 
--- a/test/setup.js
+++ b/test/setup.js
@@ -18,3 +18,13 @@ global.requestAnimationFrame = cb => cb();
 
 // Only relevant for bindGlobalEventListeners.test.js
 Object.defineProperty(window.navigator, 'platform', {value: 'iPhone'});
+
+// Prevents console from spamming test output while still allowing for debugging
+// while writting tests
+const nativeConsoleWarn = global.console.warn;
+beforeEach(() => {
+  global.console.warn = jest.fn();
+});
+afterEach(() => {
+  global.console.warn = nativeConsoleWarn;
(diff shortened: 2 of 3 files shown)


test/integration/createTippy.test.js
```diff
@@ -179,8 +179,6 @@ describe('instance.destroy()', () => {
     jest.advanceTimersByTime(500);
 
     expect(spy).not.toHaveBeenCalled();
-
-    spy.mockRestore();
```
atomiks: Why was this removed?

test/setup.js
```diff
@@ -18,3 +18,13 @@ global.requestAnimationFrame = cb => cb();
 
 // Only relevant for bindGlobalEventListeners.test.js
 Object.defineProperty(window.navigator, 'platform', {value: 'iPhone'});
+
+// Prevents console from spamming test output while still allowing for debugging
+// while writting tests
```
atomiks: `writing` spelling mistake

test/integration/createTippy.test.js
KubaJastrz: This spy is not used anymore during this test, so it's redundant.

atomiks: What if I add another spy later on in the file?

KubaJastrz: I meant this particular test case. Other `it`s will work fine.

Cleaning up is usually good even if not needed, but I haven't seen manual restoring in other test cases 🙄

atomiks: tbh I don't understand properly, but I'll trust you're right..

> I haven't seen manual restoring in other test cases

Should that be another PR?

KubaJastrz: Well, `jest` uses `spy.mockRestore()` in [their example](https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname), so I guess we can leave it there.

KubaJastrz:
I didn't make it in time with my revert commit :P
Gotta go back to work now, I will continue with tests in the evening.

PR #626, opened 2019-10-27
fix: filter out duplicate plugins

atomiks:
- Addons were `.concat`ing `defaultProps.plugins` unnecessarily since `tippy()` does it.
- Duplicate plugins should be filtered out at `createTippy()` for extra safety.

Most exported plugins when duplicated don't cause issues but make performance worse.

`unique()` could use `[...new Set(plugins)]` which is faster I think, but I'm not sure of how well-supported it is, best to use this one?

### DIFF
--- a/src/addons/createSingleton.ts
+++ b/src/addons/createSingleton.ts
@@ -24,7 +24,7 @@ export default function createSingleton(
     );
   }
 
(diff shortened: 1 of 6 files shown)


atomiks:
Also something odd is the linter is not longer complaining about lack of return type for callback functions like `.filter()` (which is way less noisy and better, maybe we can remove all of them in another PR).

Issue #623, opened 2019-10-26
v5: How to show tip on selected text?


tedypranolo:
I am trying to show a tip exactly on the selected textnode instead of the parentElement. What is the proper way to do this with v5? (If possible with sticky:true).
I have tried passing a placeholder element and overriding `getClientBoundingRect `with `window.getSelection().getRangeAt(0).getBoundingClientRect` but it doesn't seem to work.

KubaJastrz:
Interesting problem. Here's something I threw together. Use at your own risk 😉 

https://codepen.io/hawking/pen/NWWgwWr

atomiks:
@KubaJastrz there seems to be a bug when clicking off the selection again, the tippy "re-shows" when the selection goes away. Can be fixed by wrapping the `mouseup` logic in `setTimeout` but there's prob a better solution there.

Also would benefit from `inlinePositioning` but that uses `onTrigger` and bypasses `.show()` (which has exposed a bug, actually).

tedypranolo:
This works well. Thank you @KubaJastrz and @atomiks

z0xyz:
@atomiks Is @KubaJastrz's solution is the only viable solution for this use-case?

PR #630, opened 2019-10-29
fix(inlinePositioning): set ReferenceObject in onShow lifecycle hook

### LINKED ISSUE #627: `inlinePositioning` does not work with show() method
## Bug description

`inlinePositioning` uses `onTrigger` which `.show()` method bypasses, so it doesn't work.

## Reproduction

https://codepen.io/atomiks/pen/dyyReGM

atomiks:
Fixes #627

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -757,22 +757,15 @@ export default function createTippy(
       );
     }
 
-    if (instance.popperInstance) {
-      setFlipModifierEnabled(
-        instance.popperInstance.modifiers,
-        instance.props.flip,
-      );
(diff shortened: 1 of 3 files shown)


atomiks:
Though, still doesn't work with @KubaJastrz's selection technique in #623.

Somehow it should work with this:

```js
instance.popperInstance.reference = selection.getRangeAt(0);
```

amanagr:
> instance.popperInstance.reference = selection.getRangeAt(0);

This doesn't work. ` instance.popperInstance` is `null` in `onShow` function. It is defined in `onShown` but anyway, the intended purpose is not served.

Issue #631, opened 2019-10-30
Typing problem with new native props + plugins


atomiks:
A user created a plugin with name `x` that takes a `number`:

```js
tippy(reference, {
  x: 1000,
  plugins: [x]
});
```

We introduce a new native prop with name `x` that takes a `string`

```js
tippy(reference, {
  x: 'content'
})
```

```ts
interface Props {
  // ...
  x: string;
}
```

With a runtime bailout check we can prevent errors and add a warning that their plugin now conflicts with a native prop (I expect this to be very rare):

```js
bailIfPlugin(plugins, 'x', () => {...});
```

But for types, there is a problem. If they weren't typing their plugin prop, they will now get an error since `Props` now has `x` types as `string`, but they are passing in `number`.

Is it fine that types can suddenly emit errors/warnings upon minor updates? Or do we need to remove `[key: string]: any` and force strict types (which will still emit errors if they weren't strict typing their plugin prop, but only for that release, not future ones that introduce new props)?

 Looking at the DefinitelyTyped repo, changes in types happen frequently for correctness between minors and aren't considered breaking changes. Even TS itself introduces breaking in minors.

/cc @KubaJastrz

KubaJastrz:
I'm in favor of stricter types. Maybe TypeScript itself doesn't encourage strict typing by default (exhibit [1](https://twitter.com/sindresorhus/status/1118025528894611456) and [2](https://github.com/Microsoft/TypeScript/issues/30825#issuecomment-523668235)), it is a good idea to promote strictness wherever possible (exhibit [1](https://twitter.com/jessidhia/status/1097845261416554496)). If you don't want types, you can just use plain JS.

atomiks:
I think all interfaces that use `Props` need to become generic? That interface is used to type `LifecycleHooks`, `Instance`, etc.

```ts
// Haven't tested this example

const tippyExtended: Tippy<Props & FollowCursorProps> = (targets, props) => {
  return tippy(targets, props);
}

const instance = tippyExtended(reference, {
  followCursor: true,
  // This would need to take Instance<Props & FollowCursorProps>
  onShow(instance) {}
}); // Instance<Props & FollowCursorProps>

// type error
instance.setProps({followCursor: 'invalid'});
```

PR #632, opened 2019-11-01
types: refactor interfaces

atomiks:
- Removes `[key: string]: any` from `Props` (see #631)
- Makes all interfaces that use `Props` become generic (to use `TProps`)
- Add included plugins to `Props` so typechecking is enabled by default even if the plugin isn't passed (there is a DEV warning for this)

### DIFF
--- a/src/addons/createSingleton.ts
+++ b/src/addons/createSingleton.ts
@@ -1,4 +1,4 @@
-import {Instance, Props, Plugin} from '../types';
+import {Instance, Props, CreateSingleton} from '../types';
 import tippy from '..';
 import {preserveInvocation, useIfDefined} from '../utils';
 import {defaultProps} from '../props';
@@ -8,12 +8,12 @@ import {throwErrorWhen} from '../validation';
  * Re-uses a single tippy element for many different tippy instances.
  * Replaces v4's `tippy.group()`.
  */
-export default function createSingleton(
-  tippyInstances: Instance[],
-  optionalProps: Partial<Props> = {},
+const createSingleton: CreateSingleton = (
+  tippyInstances,
+  optionalProps = {},
   /** @deprecated use Props.plugins */
-  plugins: Plugin[] = [],
-): Instance {
+  plugins = [],
+) => {
   if (__DEV__) {
     throwErrorWhen(
       !Array.isArray(tippyInstances),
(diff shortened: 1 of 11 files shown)


atomiks:
Also added: function overloading types so TS knows the return value based on the `targets` argument:

```js
tippy('button') // Instance[];
tippy(document.body) // Instance;

delegate('button', {target: 'string'}) // Instance[];
delegate(document.body, {target: 'string'}) // Instance;
```

atomiks:
Example extender function with `followCursor` plugin:

```ts
import tippy, {
  Props,
  FollowCursorProps,
  followCursor,
  SingleTarget,
  Instance,
  MultipleTargets,
  Targets,
} from 'tippy.js';

type ExtendedProps = Props & FollowCursorProps;

// 1. How do I pick the call signature from `Tippy` only (ignore statics)?
// Instead of redefining it here:
interface Tippy<TProps = Props> {
  (targets: SingleTarget, optionalProps?: Partial<TProps>): Instance<TProps>;
}
interface Tippy<TProps = Props> {
  (targets: MultipleTargets, optionalProps?: Partial<TProps>): Instance<
    TProps
  >[];
}

const tippyExtended: Tippy<ExtendedProps> = (
  targets: Targets,
  optionalProps: Partial<ExtendedProps> = {},
) => {
  // @ts-ignore 2. How to remove this?
  return tippy(targets, {
    ...optionalProps,
    plugins: [followCursor, ...(optionalProps.plugins || [])],
  });
};

// Typechecking for these lines works properly
const instance = tippyExtended(document.body);
const instances = tippyExtended('button');
instance.setProps({followCursor: 'invalid'});
instances.forEach(() => {});
```

There are two issues I encountered, I've labeled them with comments.

atomiks:
I think we should just add all the plugin types to `Props` itself 🤷‍♂

For custom plugins, the user does something like:

`src/tippy.ts`

```ts
import tippy, {Tippy, Props, Plugin} from './types';

interface XProps {
  x: boolean;
}

interface YProps {
  y: number;
}

export const x: Plugin<XProps> = {
  name: 'x',
  defaultValue: false,
  fn: () => ({}),
};

export const y: Plugin<YProps> = {
  name: 'y',
  defaultValue: 0,
  fn: () => ({}),
};

type ExtendedProps = Props & XProps & YProps;

// Plugins don't get passed automatically, but there is typechecking for the props.
// The DEV warning lets them know if they didn't pass the plugin object.
export default (tippy as unknown) as Tippy<ExtendedProps>;
```

`src/index.ts`

```ts
import tippy from './test';

// Both overloading and typechecking for custom plugins work
const [i] = tippy('button', {x: true, y: 1000});
i.setProps({x: false});

// Also included plugins become part of `Props` itself
i.setProps({followCursor: true}); // DEV warning that they didn't pass the plugin
```

This is the simplest usage I think.

Issue #649, opened 2019-12-03
Possible unexpected behaviour with trigger: 'click mouseenter'


mattsbennett:
## Bug description

As in the title, I _think_ there might be unexpected behaviour with `trigger: 'click mouseenter'`. It also could be that I have expectations that don't match the intent of the existing option interactions!

## Reproduction

Very simple case: https://codepen.io/mattsbennett/pen/XWJJNqr

Here's what I would expect:

- Action: Hover element.
    - Result: Tip shown (expected)
- Action: Element still hovered. Now, click tip.
    - Result: Tip hidden (unexpected)
    - Expected Result: Tip remains shown
- Action: Element still hovered. Click tip again.
    - Result: Tip shown (expected)
- Action: Move cursor so element no longer hovered.
    - Result: Tip hidden (unexpected)
    - Expected result: Tip shown until clicking outside element (as per default `hideOnClick` and `trigger: 'click'`).

In other words, if `click` _and_ `mouseenter` are specified in `trigger` the tip should (I think) remain shown on clicking the element and until the element is clicked outside (as per `hideOnClick`) (instead of the existing toggle-on-click and always-hide-on-mouseleave behaviour).

So, `mouseenter` and `mouseleave` would toggle the tip until the element is clicked, after which only clicking outside the element would close it (assuming the default `hideOnClick: true`). I believe this would be more consistent with the behaviour of `trigger: 'click'` on its own, and still respect the behaviour of `trigger: 'mouseenter'`.

atomiks:
That kind of behavior is somewhat ambiguous. I don't think it makes sense to have `mouseenter` and `click` in the same trigger in the first place? 🤔 What is the use case for it?

mattsbennett:
Gotcha - I didn't realize the intent was that these two triggers would not be used together. If that's the case, perhaps a clarification in the docs might help? That said, here's my use-case:

For interactive tooltips/popovers that contain buttons/other interactive UI elements, having to hover the trigger element, then carefully guide the cursor over top of the tip to get to the other elements is less than desirable UX. The proposed behaviour above would provide the following benefits:

- `mouseenter` would aid in discoverability of the interactive tip (currently `click` triggered tips are not very discoverable for use in any common user-flow).
- Having the `click` toggle the tip open until clicked outside would resolve the cursor-guiding requirement mentioned above.

atomiks:
[...]

PR #659, opened 2019-12-15
Modify trigger: 'mouseenter click' behaviour.

mattsbennett:
This is an implementation proposal for the changes required for the behaviour change proposed at https://github.com/atomiks/tippyjs/issues/649. Currently, the `mouseenter` behaviour always overrides the `click` behaviour. At a high-level, the behaviour change here amounts to respecting the `mouseenter` behaviour only _until_ the reference element is clicked. Then, the existing click behaviour is respected, until the tippy is hidden (e.g. as per `hideOnClick` = `true` || `'toggle'`). Once hidden again, the existing `mouseenter` behaviour is respected once more.

Lower-level description:
- When both mouseenter and click `trigger`s are specified, respect existing mouseenter behaviour until reference element is clicked.
- On click, show the tippy and no longer hide on mouseleave.
- Respect all pre-existing `hideOnClick` behaviour (true, false, 'toggle').
- ~Only toggle the tippy on clicking the reference element when hideOnClick='toggle' (currently the tippy is _always_ toggled on click (unless `hideOnClick` is set to `false`).~ This conflicts with the preceding point, hence removed (see discussion below).

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -82,7 +82,7 @@ export default function createTippy(
   let hideTimeout: any;
   let scheduleHideAnimationFrame: number;
   let isBeingDestroyed = false;
-  let tipIsVisibleFromClick = false;
+  let isVisibleFromClick = false;
   let didHideDueToDocumentMouseDown = false;
   let popperUpdates = 0;
   let lastTriggerEvent: Event;
@@ -310,7 +310,7 @@ export default function createTippy(
     }
 
     if (instance.props.hideOnClick === true) {
-      tipIsVisibleFromClick = false;
+      isVisibleFromClick = false;
       instance.clearDelayTimeouts();
       instance.hide();
 
@@ -449,8 +449,7 @@ export default function createTippy(
     // Toggle show/hide when clicking click-triggered tooltips
     if (
       event.type === 'click' &&
-      (!includes(instance.props.trigger, 'mouseenter') ||
-        tipIsVisibleFromClick) &&
+      (!includes(instance.props.trigger, 'mouseenter') || isVisibleFromClick) &&
       instance.props.hideOnClick !== false &&
       instance.state.isVisible
     ) {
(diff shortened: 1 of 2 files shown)


atomiks:
Thanks for the PR! Sorry for the delay, I will try and check this tonight.

atomiks:
This is almost good, just `hideOnClick: true` (default) doesn't hide the tippy on the second click. Then it should be toggle-able.

`hideOnClick: "toggle"` just means it can be toggled by clicking the ref element, but not clicking outside. In this context they should work the same.

mattsbennett:
Ah, right - thank you for the feedback! I'll get back to this ASAP (probably the weekend).

mattsbennett:
Alright, I've addressed the `hideOnClick: true` issue @atomiks !

atomiks:
`hideOnClick: true` still isn't toggling for me with the latest changes? I just realized this sounded ambiguous:

> This is almost good, just hideOnClick: true (default) doesn't hide the tippy on the second click.

What I meant was this was the behavior with the PR, not what it should be. 😅 

`hideOnClick: true` and `hideOnClick: "toggle"` should work the same when clicking the reference element while hovering over it. The difference between them is that `true` allows clicks outside, while `toggle` doesn't.

src/createTippy.ts
```diff
@@ -82,6 +82,7 @@ export default function createTippy(
   let hideTimeout: any;
   let scheduleHideAnimationFrame: number;
   let isBeingDestroyed = false;
+  let tipIsVisibleFromClick = false;
```
atomiks: I think `isVisibleFromClick` (without `tip`) is fine

mattsbennett:
Thanks for the review! I've made that renaming change.

atomiks:
Thanks a lot!

mattsbennett:
You bet!

PR #663, opened 2019-12-20
feat: transitionDimensions plugin [wip]

atomiks:
This transitions the dimensions of the tippy smoothly in size when its content is updated. This makes singleton transitions look a lot better.

Content inside should be able to be transitioned by the user by making an absolutely-positioned cloned node of the previous content in `onBeforeUpdate`, then transitioning it out in `onAfterUpdate`.

Unfortunately there is a bug in Safari regarding early cancellations of the transition. When cancelled early, the dimensions don't get updated to the next content's dimensions and stay at the dimensions they were when it got cancelled. Any help figuring that out would be great. 

No idea about IE11 but probably just don't bother with these effects on such an ancient browser 🤷‍♂

### DIFF
--- a/build/base-iife.js
+++ b/build/base-iife.js
@@ -5,10 +5,17 @@ import animateFill from '../src/plugins/animateFill';
 import followCursor from '../src/plugins/followCursor';
 import inlinePositioning from '../src/plugins/inlinePositioning';
 import sticky from '../src/plugins/sticky';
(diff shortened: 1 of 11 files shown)


atomiks:
Safari bug fixed.

However, when the tippy is centered, the arrow's transition on Safari looks horribly jerky 😭

PR #667, opened 2019-12-31
Prevent Object [object Performance] has no method 'now' issue

CPatchane:
Browsers on Android <= 4.4 seems to not have `now` property on the `window.performance` object, this PR should fix that.

Original issue:
```
TypeError
Object [object Performance] has no method 'now'
```
Screenshot 2019-12-31 at 10 52 59

### DIFF
--- a/src/bindGlobalEventListeners.ts
+++ b/src/bindGlobalEventListeners.ts
@@ -17,7 +17,7 @@ export function onDocumentTouchStart(): void {
 
   currentInput.isTouch = true;
 
-  if (window.performance) {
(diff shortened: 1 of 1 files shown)


atomiks:
I would recommend polyfilling this if you need to, as I've dropped official support for ancient browsers like Android 4.4 since 5.0 - according to `caniuse` it's only at 0.27% usage. The CSS files do not have any `-webkit-` prefixes for example (although if importing, your autoprefixer handles it, but the point stands for basic/CDN usage etc.)

PR #683, opened 2020-01-16
update readme.md

cocoa-maemae:
It looks like there is also tippy adapted to vuejs
I added link to https://github.com/KABBOUCHI/vue-tippy


atomiks:
Hi, thanks for the PR, but only official wrappers will be added to the README. Only React is officially supported right now.

PR #689, opened 2020-01-27
 Add support for 'focusin' trigger to resolve #682 

### LINKED ISSUE #682: Add a focusin trigger
Currently, the only triggers are `mouseenter`, `focus`, `click`, and `manual`. This means you have to directly wrap the target element (or reference it, which can be difficult for React apps). It would be nice to have a way to show the tooltip when focus enters the target element and hide it when focus leaves.

iansan5653:
Adds support for an additional trigger called `focusin` that fires when focus _enters_ the target element, rather than when the target element is focused. Also adds this value in the documentation and adds tests for it. 

One additional small modification is to fire the `mouseEnter` event before doing the bubbling test on line 155 of `createTippy.test.js`, so it's not affected by previous tests.


atomiks:
Thanks!

Issue #616, opened 2019-10-23
Improve CI infrastructure


atomiks:
Remaining:

- Type tests
- Build tests (`build/` files), especially the UMD versions

KubaJastrz:
Sounds like fun. I can definitely take a look on automating those processes.

Maybe we can even use GitHub Actions for this. I've only used Jenkins so far, so this will be interesting to learn.

atomiks:
Travis seems to have disappeared on GitHub UI but it's still running in the background

atomiks:
Travis disappeared yet again 🙄

atomiks:
FYI I've added image snapshot tests for the `next` branch (thanks to @FezVrasta I copied lots of his setup 🤣) however they're failing on CI by a few 0.3%-5% and I don't know why. I think it might be due to the text rendering on Linux vs Mac, but at the same time, I kind of need to test text color for the themes... I don't know how to access the diffs without uploading them to a service either to see why they're actually different, but it's probably the text. 

I think we had that issue on Popper and we had to make it transparent.

FezVrasta:
You may setup a docker image to run the tests locally, so that the snapshots are identical on CI and locally.

atomiks:
I worked it out — the button styling on Linux is different, I used `all: unset` and they're passing now. The docker thing would be good but too complex for me to set up at the moment :-\

atomiks:
@KubaJastrz doesn't seem like `release: v6.0.1` triggered a docs build?

KubaJastrz:
@atomiks, it did. But you haven't provided the `GITHUB_TOKEN` variable in [travis ui settings](https://travis-ci.com/atomiks/tippyjs/settings). 
https://travis-ci.com/atomiks/tippyjs/jobs/295120942

atomiks:
Fixed 👀

atomiks:
Most of this is done now, can track the remaining work in the type test PR

PR #692, opened 2020-02-01
chore: deploy docs with travis

KubaJastrz:
Add stage to Travis CI that builds and deploys docs website to github pages. Runs only on `master` branch.

Related: #616


KubaJastrz:
@atomiks, you may need to provide `GITHUB_TOKEN` variable, like described in [Travis docs](https://docs.travis-ci.com/user/deployment/pages/#setting-the-github-token).

atomiks:
One problem is pull requests changing the docs along with source code. They will reflect too new a version before it's released (which may be days). Is there a way to detect if changes to the docs are a result of a new feature committed to `src/` or something and postpone those until after a release?

KubaJastrz:
I think the easiest solution would be to have stable `master` branch and `dev` for unreleased features. Docs would be only built for `master`, or potentially, we could have separate version with something like netlify/now.sh preview for `dev`.

Other than that, maybe tagging commits with `docs: <message>` could trigger such deploys (with manual deploy also available in case something goes wrong).

atomiks:
> I think the easiest solution would be to have stable master branch and dev for unreleased features. Docs would be only built for master, or potentially, we could have separate version with something like netlify/now.sh preview for dev.

I only see repos merging PRs into `master` so this may be confusing.

> Other than that, maybe tagging commits with docs: <message> could trigger such deploys (with manual deploy also available in case something goes wrong).

I like this I think. I'd like it to deploy in two cases:

- docs commits (as you said `docs: commit message`)
- release commits (maybe a `release` prefix or something? `release: 5.1.2`)

Is it possible?

KubaJastrz:
> I only see repos merging PRs into master so this may be confusing.

If we'd set `dev` as default branch then I think this is manageable. Although that would mean `dev` is displayed when you browse the repository.

> Is it possible?

I think so, Travis seems to have `TRAVIS_COMMIT_RANGE` env variable that we can use to scan for commit messages.

KubaJastrz:
I've made some progress, but it looks like `TRAVIS_COMMIT_RANGE` will contain wrong commit hashes in case of rewriting history (force push).

If we don't force push to the `master` branch we should be fine 😄 Otherwise, I have to default to not deploying docs.

website/scripts/should-deploy-docs.js
```diff
@@ -0,0 +1,49 @@
...
+console.log(`TRAVIS_COMMIT_MESSAGE=${TRAVIS_COMMIT_MESSAGE}`);
+
+let commitMessages;
+
+if (TRAVIS_COMMIT_RANGE) {
+  // see https://github.com/travis-ci/travis-ci/issues/4596
+  const COMMIT_RANGE = TRAVIS_COMMIT_RANGE.replace('...', '..');
+  const BASE_COMMIT = COMMIT_RANGE.split('..')[0];
+
+  const baseCommitType = execSync(`git cat-file -t ${BASE_COMMIT}`)
```
KubaJastrz: I need to add try-catch for this or add `|| true` at the end of command, since it will error out anyway...

website/scripts/should-deploy-docs.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+local NODE_OUPUT=$(node ./scripts/should-deploy-docs.js)
```
KubaJastrz: I've tried, but working in bash is a nightmare. So the main logic is hidden in node 😅

KubaJastrz: Now that I think about it, maybe this bash script is redundant. I've added it to export `SHOULD_DEPLOY_DOCS` variable to pipeline, but it's only used to terminate the pipeline (which is being done here and could be moved to JS file as well).

atomiks:
I have no idea about any of this, does it work? 😄

KubaJastrz:
Ok, so here's how it works:

1. Regular CI pipeline stays the same, it will still implicitly run `npm ci` and `npm test` on node versions specified in the [`node_js` matrix](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/).

2. There's new job in the pipeline that runs if `branch = master` that is responsible for building and deploying gatsby website to [`gh-pages` branch](https://docs.travis-ci.com/user/deployment/pages/). That condition also prevents it from running on PR branches.

3. To make sure that deploy job is being run only once (Travis defaults to running all jobs on all configurations in [build matrix](https://docs.travis-ci.com/user/build-matrix/)), there's explicit version set: `node_js: 12`.

4. To make sure that docs deployment is happening only when conditions are being met (commit message starts with either `docs: ` or `release: `), on [`before_install` lifecycle](https://docs.travis-ci.com/user/job-lifecycle/) we run validation script that decides to proceed or terminate the job.

5. As I'm not so comfortable with bash scripts, I've moved main functionality to js script and used some stdout shenanigans to read the output.

6. I've taken an advantage of [`TRAVIS_COMMIT_RANGE` env variable](https://docs.travis-ci.com/user/environment-variables/) which allows us to read messages of commits affected by the build. Unfortunately it's [a little flawed](https://github.com/travis-ci/travis-ci/issues/2668) when it comes to force pushes that rewrite git history, hence the `git cat-file` check and early exit (taken from [a handy article](https://dev.to/ahferroin7/skip-ci-stages-in-travis-based-on-what-files-changed-3a4k)).

atomiks:
Thank you for the hard work 🎉  

However, I realized another flaw: when we make other changes to the docs (`docs: ...`) after a source code + doc change, but before a release, we have the original problem again. 

I have an idea:

All changes to docs along with source code should be commented out. 

e.g for `trigger` with the v5.2.0 feature:

```html
Possible values: "mouseenter", "focus", <!-- unreleased: "focusin", --> "click".
```

When we perform a `release:` commit, it should uncomment out any MDX comments with the `unreleased` prefix. Possible?

---


Also, I think there is an issue with Gatsby chunks with deploys, we need to [cache the `public/` and `.cache` directories](https://github.com/popperjs/popper-core/blob/master/.travis.yml#L4-L10) to preserve old chunks when doing new deploys (so stale users don't have a broken experience): https://github.com/gatsbyjs/gatsby/issues/18866 (ideally Gatsby would do a full page reload on 404 chunk but it doesn't right now).

KubaJastrz:
> When we perform a release: commit, it should uncomment out any MDX comments with the unreleased prefix. Possible?

We are now entering the babel plugins category. Should be possible.

Although I really think that the better solution would be to have two separate environments / some other way of versioning the docs.

atomiks:
> We are now entering the babel plugins category. Should be possible.

Won't a quick global `str.replace()` regex do the job? I always do that usually for these sorts of tricks, and have never written a Babel plugin that actually gives you the AST 🤣

KubaJastrz:
> Won't a quick global str.replace() regex do the job?

Well, maybe. But what about adding or removing pages for the new release? Or reordering indexes/renaming them for some reason? Not sure that will work so smoothly :/

atomiks:
You're right, I don't expect that to be too common, but it's still another flaw with the approach. A separate branch is the only sane solution I guess, even though I don't see it much on GitHub open source repos 🤔

KubaJastrz:
@atomiks, how do we proceed with this now that we have separate directories for major versions?

atomiks:
@KubaJastrz I'm not sure what you mean, it should just work without any special handling?

KubaJastrz:
Well, yes. Do you still want the check for `docs` and `release` commit message prefixes and if so, can we continue with this PR?

atomiks:
Yes, sure. It's not perfect but probably better than most other techniques, aside from the separate branch

atomiks:
Is this good to go @KubaJastrz?

KubaJastrz:
Should be fine

PR #694, opened 2020-02-04
Hide tippy on focusout from child element to fix #693

### LINKED ISSUE #693: Tippy with "focusout" trigger doesn't disappear when focus leaves the container from a child element
When I closed #682, I used the same `onBlur` handler that was already use, and just called it on `focusout` in addition to `blur`. That handler checks whether `event.target` is the target element for the Tippy instance and if it is, hides the tooltip. This ensures that the tooltip isn't hidden if focus leaves the target element and enters the tooltip.

This appears to work at first glance, because when `focusout` is dispatched by the container element, the `event.target` is the container. However, when focus leaves the container element from a child, `event.target` is that child. 

For example, when the user tabs through the following document, `focusout` will be dispatched by `#container` when focus goes from `#b` to `#c` as expected. However, the `target` of that event will be `#b` _not_ `#container` as I expected:
```html
<a href="#" id="a">a</a>
<span id="container">
  <a href="#" id="b">b</a>
</span>
<a href="#" id="c">c</a>
<a href="#" id="d">d</a>
```

Thus, Tippy thinks that focus has gone to the tooltip and so the tooltip should stay visible, where in fact it has gone to the next element and the tooltip should hide.

iansan5653:
This resolves the problem with the tooltip not disappearing when focus leaves the element. I was unable to write a test for this as I don't think it's possible to fire an event on one element while having the `target` of that event be a different element, but I could be wrong.

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -422,10 +422,10 @@ export default function createTippy(
           on('mouseleave', onMouseLeave as EventListener);
           break;
         case 'focus':
-          on(isIE ? 'focusout' : 'blur', onBlur as EventListener);
+          on(isIE ? 'focusout' : 'blur', onBlurOrFocusOut as EventListener);
           break;
         case 'focusin':
-          on('focusout', onBlur as EventListener);
+          on('focusout', onBlurOrFocusOut as EventListener);
           break;
       }
     });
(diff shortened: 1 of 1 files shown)


atomiks:
Thank you

atomiks:
Just realized IE11 uses `focusout` for whatever reason (missing `event.relatedTarget`?)

I think there will be a problem with it 😭

Issue #720, opened 2020-03-03
Simplify styles definition


woothu:
## Problem

Styles of tooltip could be easier to define.

## Solution

I think that the '.tippy-box' element could be responsible for styling tooltip. 
I don't have insights into style nuances, but I think it should be manageable to set styling based on this one element parameters.

I have made a demo fiddle which partially shows what I mean by simplifying style definition:
http://jsfiddle.net/zu7ewLsx/19/

One thing I didn't get on fiddle is proper arrow background-color. I wanted to get the following result, which demand ::before element to inherit background-color and ::after element to inherit border-color

```css
.tippy-box[data-theme~='ugly-example'] {
  border-color: green;
  text-color: black;
  background-color: dodgerblue;
}
```
Result: 
![image](https://user-images.githubusercontent.com/32914662/75791947-9a506580-5d6d-11ea-8dd3-be0325c65290.png)

Advantages:
- super easy theme API,
- super light themes,
- if 'styleClasses' prop was added to API (string with classes passed to tippy-box), user can define styles by existing classes, which for now is not possible i.e 
```js
{
  styleClasses: 'user-text-color-class user-background-color-class user-border-class'
}
```

Disadvantages:
- might be breaking change (Although it is possible that current selectors are more specific and no breaking change would occur)
- SVG arrows might need JS handling to also have this behavior

Overall I think that the idea of manipulating the whole design of tooltip just by these three parameters is simplifying the API greatly :)

atomiks:
A PR implementing this would be great assuming it's fully backwards-compat.

woothu:
Do you mean backwards compatibility with docs examples or any possible styles?

atomiks:
I mean if a user updates the version, and their styles break, it's not backwards-compat

woothu:
It is possible to only cover some possible users theme usages making any change on CSS

woothu:
solved with #727

PR #725, opened 2020-03-05
feat: add arrowInheritsStyles prop

woothu:
#720 solution that causes no breaking changes.

Considerations: 
- Style tag creation would be not necessary, but 'css inheritance way' demand CSS rewrite (breaking changes - possibility for version 7),
- JavaScript handling is inevitable if we want to style arrow based on box background-color,
- Performance measured from 6milisecons for first and 1ms for later calls of styleArrow(), first call time is caused by style tag creation, it could be created beforehand for optimization

Demo:
1.   set 'arrowInheritsStyles: true'
2. add
```
.tippy-box[data-theme~='light-border'] {
  border-color: teal;
  background-color: wheat;
}
```
result:
image

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -1,3 +1,4 @@
+import {ARROW_CLASS} from './constants';
 import {createPopper, Modifier} from '@popperjs/core';
 import {currentInput} from './bindGlobalEventListeners';
 import {isIE, isIOS} from './browser';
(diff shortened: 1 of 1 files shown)


atomiks:
Thank you for the PR 👋, however I don't think I'll add something like this to this library because you can implement this functionality as a plugin since it's not accessing any internal code, it looks purely to be accessing public APIs!

woothu:
Would you publish such a plugin in source code and add it to the docs?

Issue #726, opened 2020-03-05
Add transparent ::after element to default css 


woothu:
## Problem

Users cannot easily add arrow border because ::after element is only present in 'light-border' theme 

## Solution

Adding ::after element with transparent borders to default styles

woothu:
closed with #727

PR #727, opened 2020-03-06
feat: add transparent ::after element, refactor styles

woothu:
I was able to solve  #726 and #720 without (or very unlikely) breaking changes!

Compatibility:
Styles should remain the same due to:
  - ::before element now inherits from text color of arrow, but it was probably never changed by users. (No usage)
  - added ::after element which is transparent as long as border-color of arrow element is not changed, which is also unlikely to be changed by users, since it had no effect on the styles
  - even if these changes were made, themes have priority through stronger CSS selectors
  
There is a probability of a break in custom themes (i.e if someone made similar changes I did), but I really doubt it.

Feature: 
basic styles + added user CSS:
```
.tippy-box {
  border: darkcyan solid 1px;
  background-color: lightblue;
}
.tippy-arrow {
  border-color: darkcyan;
  color: lightblue;
}
```
You have following effect
image

Without border:
```
.tippy-box {
  background-color: lightblue;
}
.tippy-arrow {
  color: lightblue;
}
```

image

Btw. This feature highly improves #725 plugin because it is no longer necessary to create style tag.


atomiks:
> without (or very unlikely) breaking changes!

screenshot showing failing tests

😩

woothu:
Tests fails after color change from '$color' to 'initial' on sides (and giving .tippy-arrow color: $color)

And it shows '42px' difference on a tippy

imho its tool bug, because a change of the color, cannot change tooltip size (as far as I understand CSS 😂)

image

#Find 5 differences :D

atomiks:
The image snapshots are showing the arrow is invisible entirely.

atomiks:
I think it's cause of this stylesheet that gets injected into the tests lol:

```css
* { 
  color: transparent !important; 
}
```

I think that's obscure enough not to be considered a problem, I can adjust it

woothu:
Strange, in my case snapshot arrows were visible

woothu:
I uploaded snapshots for review

atomiks:
For me, and in CI, it shows the arrow is invisible like due to the rule I posted above (inherits transparent color): https://github.com/atomiks/tippyjs/blob/ae56081e1b8871de9f1886863a94a54bdca6192c/test/utils.js#L41-L43

I can't remove that though because of text rendering differences between CI and Mac unless I set up a docker container 😵

woothu:
image

Now I see two options:
- you don't want text display to be tested, then '!important' deletion would be enough (or maybe just target .tippy-content)
```
     * { 
       color: transparent; 
     } 
```
- you want also include font style to tests, then snapshots should be updated

atomiks:
> you don't want text to be tested, then '!important' deletion would be enough (or maybe just target .tippy-content)

Can you try without `!important` (delete it in the code I linked)? `.tippy-content` isn't enough because there's other text in headings, reference buttons, etc

woothu:
Seems like everything is working :)

atomiks:
How many bytes does this add to the core CSS? Also can we add some screenshot tests for this (border styles inheritance)?

woothu:
It adds 79bytes gzip, I will consider it negligible concerning bloat, please have in mind that development speed and code clarity are very important for developers. This is basic functionality, not a theme so it shouldn't be imported separately imho.

But concerning corner-case breaking changes, it is reasonable to make after element a separate import until 7.0 version. 

It will also allow making .tippy-arrow .border-color 'inherit' instead of 'transparent' which further simplifies API (user doesn't have to specify arrow border color if it is identical to box border color). This is breaking change for users who have border on the box, but not on the arrow.

woothu:
Concerning the feature of inheriting border-color from box by ::after element, to make it work properly, in after.css box border-color is set to transparent by default (to avoid default border on ::after element if border on the box is not specified)

woothu:
I wanted to give box border solid style by default in after.scss, but for some reason, it breaks the functional tests :( For now, I left it without default border-style.

atomiks:
This seems good now, but one thing that's a bit uncertain with this is SVG arrow borders. In the `light-border` theme, I have to use a base-64 encoded SVG to place as the border, and it's entirely dependent on the type of shape used (which can be anything). Can you think of a decent solution for that?

woothu:
SVG should contain another path which is the border of it, there is no other good way of achieving border. I have tried using SVG "stroke" but it won't work well. (Because stroke is added to the whole shape)

image

Each path can have its class which would be targeted in user themes. (like .svg-arrow-content,  .svg-arrow-border)

atomiks:
Can they do this:

```html
<svg class="tippy-svg-arrow--fill">
  <path ... />
</svg>
<svg class="tippy-svg-arrow--border">
  <path ... />
</svg>
```

This would allow them to put a class on each SVG element to style.

```js
arrow: svgArrowHTML
```

woothu:
It is responsibility od SVG file to deliver border imho, moreover if svgs/paths would have class names then CSS selectors would be easier to read i.e.
```
.svg-arrow-body{}
.svg-arrow-content{}
```

It is better also to not just mimic border, but give it bigger shape, so additional positioning is not needed. The easiest way is adding stroke probably.

atomiks:
The svg file is agnostic about the shape, so I think it should belong in this `after` CSS file which is what provides a separate border for arrows — the user could override the `background-image: url(...)` with a shape that matches their `arrow: svgArrowHTML` but with the rest of the styles in place. 

And maybe the file should be called `arrow-border` instead of `after`

woothu:
I don't know what you mean by saying SVG is agnostic about shape. Path determines the shape of the SVG, and another path could determine shape of the border...

atomiks:
I think the `nth-child()` technique seems simplest as seen in the CodePen. The first child SVG is the border, which is 1px offset from the fill SVG (the second child).

atomiks:
I'll see what I can do for the SVG. Thanks for this PR!

woothu:
Here is what I have meant:
https://codepen.io/woothu/pen/rNVdGbJ

PR #733, opened 2020-03-12
Handle onClickOutside props for tippy instance

CPatchane:
This new props allows to pass a new listener when the user click outside of a shown tippy instance.
Related to https://github.com/atomiks/tippyjs-react/issues/166

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -312,6 +312,8 @@ export default function createTippy(
       ) {
         return;
       }
+    } else {
+      instance.props.onClickOutside(instance, event);
     }
 
     if (instance.props.hideOnClick === true) {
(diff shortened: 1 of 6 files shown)


test/integration/props.test.js
```diff
@@ -319,6 +319,46 @@ describe('getReferenceClientRect', () => {
   });
 });
 
+describe('onOutsideClick', () => {
```
atomiks:
```suggestion
describe('onClickOutside', () => {
```

src/createTippy.ts
```diff
@@ -312,6 +312,8 @@ export default function createTippy(
       ) {
         return;
       }
+    } else {
+      instance.props.onClickOutside(instance, event);
     }
```
atomiks: What was the difference between not using the `else` branch and using it? (I noticed you said you fixed a problem with the demo, just trying to get full context)

CPatchane: That was the issue, I wasn't using it.
The difference is that if you aren't using the `else` it will be triggered also when you click on the button that opens the tippy. And it will close and reopen just after the tippy so you won't be able to close the Tippy by clicking again on the opener.

atomiks:
Seems good I think =]

I think we also need an interactivity hook as well but that's not related to this 😅

PR #743, opened 2020-03-24
test: add type testing

KubaJastrz:
Added basic type checks for exported API.

### DIFF
--- a/src/props.ts
+++ b/src/props.ts
@@ -31,7 +31,6 @@ export const defaultProps: DefaultProps = {
     content: 'auto',
     expanded: 'auto',
   },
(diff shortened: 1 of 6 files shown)


src/props.ts
```diff
@@ -31,7 +31,6 @@ export const defaultProps: DefaultProps = {
     content: 'auto',
     expanded: 'auto',
   },
-  content: '',
```
KubaJastrz: duplicated by `renderProps.content`

atomiks:
Great work 🙌 should we put the file inside `test/` instead of the root?

PR #819, opened 2020-07-25
Update README.md

seb3s:
Remove useless section that currently redirect to a 404. There already is a Doc section at the top of the readme.


atomiks:
I think we should keep it but just fix the link?

seb3s:
Ok, I've kept the usage section, but made it point to getting started

seb3s:
Hi, is it ok for you ?

PR #820, opened 2020-07-25
Update package.json

seb3s:
So that the docs mention the latest published version


atomiks:
Deploys are currently done manually 😢  so this won't update it but I'll merge anyways.

PR #858, opened 2020-09-22
Adding an overload for the Targets type

callisto2410:
Hey.
Problem: Type Targets declared but not used:
screen-0

Solution: Add overload for Targets.
screen-1


callisto2410:
Failed tests.

PR #883, opened 2020-12-10
Fix bug: hovering a hiding content leads to incorrect position.

zhaoyao91:
Currently there is a bug when hovering the hiding content, it will be re-triggered but the position is wrong, since the previous cursor index is lost.

Kapture 2020-12-11 at 01 20 55

This PR fix it by retain the cursor index.

### DIFF
--- a/src/plugins/inlinePositioning.ts
+++ b/src/plugins/inlinePositioning.ts
@@ -87,6 +87,9 @@ const inlinePositioning: InlinePositioning = {
           cursorRectIndex = index > -1 ? index : cursorRectIndex;
         }
       },
+      onHidden(): void {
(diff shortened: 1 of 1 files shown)


atomiks:
I think `onUntrigger` is necessary if you, for example, use the `focus` trigger afterward. Does that still work fine with this change?

zhaoyao91:
> I think `onUntrigger` is necessary if you, for example, use the `focus` trigger afterward. Does that still work fine with this change?

Sorry for this missing. I have added `onHidden` for an alternative and tried that under `mouseenter focus` it now behaves as previous.

I am not sure is there any other missing, thanks for further checking.

PR #892, opened 2021-01-11
Add show, showNext, showPrevious methods to singletons

theo-staizen:
This PR adds the following features/changes to Singleton addon

- Refactored logic from `onTrigger` into `prepareInstance` so that it can be reused in `show`
- Fixed `show` method  for Singletons and introduced 3 overloads
  - Specific tippy instance
  - Specific reference element
  - Index number (relative to references array)
  - Show first tippy instance if no argument is passed (this was broken and showing empty tippy at top left)
- Introduced `showNext` and `showPrevious` methods on Singletons
  - Both methods will cycle once you reach last/first tippy instance
- `showOnCreate` now works correctly for singletons
- Updated `addons.mdx` docs

Our use-case for these changes: We are trying to create a guide/intro where we need to programatically show a singleton instance, but we realised that show method didn't work correctly (shows empty tooltip on top left of screen). So we fixed that and also introduced show next/previous to simplify cycling through the steps of the guide.

### DIFF
--- a/src/addons/createSingleton.ts
+++ b/src/addons/createSingleton.ts
@@ -30,9 +30,10 @@ const createSingleton: CreateSingleton = (
 
   let individualInstances = tippyInstances;
   let references: Array<ReferenceElement> = [];
-  let currentTarget: Element;
+  let currentTarget: Element | null;
   let overrides = optionalProps.overrides;
   let interceptSetPropsCleanups: Array<() => void> = [];
+  let shownOnCreate = false;
 
   function setReferences(): void {
     references = individualInstances.map((instance) => instance.reference);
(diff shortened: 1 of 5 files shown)


src/addons/createSingleton.ts
```diff
@@ -111,6 +118,51 @@ const createSingleton: CreateSingleton = (
     triggerTarget: references,
   }) as CreateSingletonInstance<CreateSingletonProps>;
 
+  const originalShow = singleton.show;
+
+  singleton.show = function(target?: ReferenceElement|Instance|number):void {
+    originalShow();
```
atomiks, line 124: Could you switch to arrow functions?

src/addons/createSingleton.ts
```diff
@@ -111,6 +118,51 @@ const createSingleton: CreateSingleton = (
...
+    return prepareInstance(singleton, (target as Instance)?.reference || target);
+  }
+
+  singleton.showNext = function():void {
+    const first = references[0];
+    if (!currentTarget) {
+      return singleton.show(0);
+    }
+    const index = references.indexOf(currentTarget);
+    singleton.show(references[index+1] || first);
```
atomiks, line 153: Have you got prettier running? There should be a space between the number and + (`[index + 1]`)

atomiks:
Thanks for this, it would be great to have some tests to cover all of this new behavior

theo-staizen:
> Thanks for this, it would be great to have some tests to cover all of this new behavior

I can't run the tests because I don't have docker on my machine

src/addons/createSingleton.ts
theo-staizen: done

src/addons/createSingleton.ts
theo-staizen: done

theo-staizen:
@atomiks i'm running into an issue with a couple of my tests.

- show method > shows the tippy instance at the given index number
- showOnCreate > shows the first tippy instance on creation

I can't debug them as I don't have docker, but both cases are working fine when testing manually.

Can u have a look too? Do I need to play with timers?

theo-staizen:
@atomiks I wrote a number of integration tests to cover the new methods introduced.
Please have a look, thanks

atomiks:
There seems to be an issue with the singleton test. The last button opens the singleton tippy at the first one. Removing `showOnCreate: true` fixes it.

theo-staizen:
@atomiks nice catch. i've added `onClickOutside` to the plugin and reset the state if it hasn't been visible yet.

Also added a test for this case

theo-staizen:
@atomiks can we please merge this?

atomiks:
Sorry, @theo-staizen will check this over the weekend and merge it!

atomiks:
Sorry for taking so long. Merging rn

PR #907, opened 2021-02-19
Make all types strictly readonly.

vilemj-Viclick:
## The problem:
It honestly started with this:
image
image

The `as const` is mandatory to keep typescript form making it an Array. It keeps the thing as a Tuple. But it also makes it readonly.
Having the API more readonly also prevents bad ideas like passing an object in and then mutating it.
## The solution:
Well just make all the types that can be made readonly `readonly`.
## Workaround:
For now I need to cast the Array like this:
image

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -12,7 +12,7 @@ export type Content =
 
 export type SingleTarget = Element;
 
-export type MultipleTargets = string | Element[] | NodeList;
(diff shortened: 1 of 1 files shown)


vilemj-Viclick:
I'm willing to fix the compilation issues. Just tell me @atomiks whether you like the direction this is taking. ;-)

atomiks:
Is this done in other libs with props?

PR #951, opened 2021-06-08
Fix: Don't stringify zIndex

bozdoz:
Not sure if this is the right place to add this change.  I noticed in tippyjs/react, if you pass:

```jsx
<Tippy
   zIndex={props.zIndex}
>
```

with `props.zIndex` being `number | undefined`, then z-index defaults to `'undefined'` if it's not passed.  Maybe this is better solved somewhere else (wherever the props are being merged).  Obviously a work-around is `zIndex={props.zIndex ?? 9999}`, but that seems bizarre for consumers to have to re-establish the API's defaults. 

Thoughts?

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -218,7 +218,7 @@ export default function createTippy(
   function handleStyles(): void {
     popper.style.pointerEvents =
       instance.props.interactive && instance.state.isVisible ? '' : 'none';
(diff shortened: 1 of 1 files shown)


atomiks:
I think this is an issue with the React wrapper, the defaults work correctly with the core lib.

PR #997, opened 2021-10-29
fix(delegate): touch input with click trigger

### LINKED ISSUE #995: tippy.delegate + trigger:"click" + hideOnClick: true = immediately hides tippy on mobile
For mobile platforms (only chrome confirmed), if you use:

```
tippy.delegate(elem, {
   target: ".thing",
   trigger: "click",
   hideOnClick: true
})
```

...Then the tippy instance briefly appears and then disappears.

Using trigger: "focus" fixes it.

I've tried setting up my own click listener in onShow to recreate hideOnClick, but again, it is immediately triggered.
I think tippy must be listening for another event (appropriate for mobile) that fires momentarily before the "click" event, but internally it's still listening for "click" for hideOnClick, so it is immediately hidden.

atomiks:
@hedgehog90

fixes #995

### DIFF
--- a/src/addons/delegate.ts
+++ b/src/addons/delegate.ts
@@ -39,7 +39,7 @@ function delegate(
 
   const nativeProps = removeProperties(props, ['target']);
   const parentProps = {...nativeProps, trigger: 'manual', touch: false};
-  const childProps = {...nativeProps, showOnCreate: true};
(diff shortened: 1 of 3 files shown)


hedgehog90:
Thanks for fixing this.
ETA for merge & release?

atomiks:
In a sec

PR #1025, opened 2022-01-05
Pass plugins to `validateProps` in `setDefaultProps`

KABBOUCHI:
Plugins are not used when validating props in `setDefaultProps`

```js
import tippy, { followCursor } from 'tippy.js';

tippy.setDefaultProps({
    followCursor: true,
    plugins: [followCursor],
})
```

```
tippy.js

`followCursor` is not a valid prop. You may have spelled it incorrectly, or if it's a plugin, forgot to pass it in an array as props.plugins. 

All props: https://atomiks.github.io/tippyjs/v6/all-props/
Plugins: https://atomiks.github.io/tippyjs/v6/plugins/

👷‍ This is a development-only message. It will be removed in production.
```


atomiks:
Thanks

PR #1026, opened 2022-01-07
Update `appendTo` docs to be inline with code/other docs

jasikpark:
https://github.com/atomiks/tippyjs/blob/60aee16c934fc861569f84a37c6a1a7f42a33fa0/src/constants.ts#L12

This is a docs change to update the description of the default setting for `appendTo`

### DIFF
--- a/website/src/pages/v6/all-props.mdx
+++ b/website/src/pages/v6/all-props.mdx
@@ -156,7 +156,7 @@ tippy(targets, {
 ### appendTo
 
 The element to append the tippy to. If `interactive: true`, the default behavior
-is `appendTo: "parent"`. See [Accessibility](../accessibility/#interactivity)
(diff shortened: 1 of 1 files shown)


atomiks:
The current docs are correct as the condition is "If `interactive: true`"

jasikpark:
Oh! I didn't see that at all 😅 It looked like the default in the text and the default in the example code described two different defaults

atomiks:
Yeah tbh the wording could be better. Like describe the actual default, then the default with the condition.

jasikpark:
I'll update my PR w/ that wording if that works! That seems like a good way to make it less confusing to me :)

jasikpark:
Only took me 7mo to update 😅

jasikpark:
anything i can do to push this PR to being merged?

Issue #1027, opened 2022-01-30
Tippy invokes `removeEventListener` and `addEventListener` with an undefined listener


LazyCompiler:
## Bug description

The events `removeEventListener` and `addEventListener` of `transitionend` and `webkitTransitionEnd` are invoked with an undefined listener.

[dom-utils.ts / line 117](https://github.com/atomiks/tippyjs/blob/21a9c842ad00cacd32c84a91080208e8c9a2eb28/src/dom-utils.ts#L117)

    ['transitionend', 'webkitTransitionEnd'].forEach((event) => {
     box[method](event, listener as EventListener);
    });

This function does not verify that "listener" is defined after being triggered from ([createTippy.ts Line 411](https://github.com/atomiks/tippyjs/blob/cac8a1260cab174d8442efa1bc1d973e298d3373/src/createTippy.ts#L411)), that also didn't although it's an global variable that being defined only after its first run:

    updateTransitionEndListener(box, 'remove', currentTransitionEndListener);
    updateTransitionEndListener(box, 'add', listener);

    currentTransitionEndListener = listener;

currentTransitionEndListener is a global var that is undefined at first load, so first call is always undefined.

## Reproduction

CodePen link: https://codepen.io/lazycompiler/pen/rNYxBjQ

atomiks:
An undefined event listener just gets ignored so I'm not sure what the bug is?

LazyCompiler:
Because this data is required and not optional (removeEventListener: [Source 1](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#parameters), [Source 2](https://www.w3schools.com/jsref/met_element_removeeventlistener.asp), addEventListener ([Source 1](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters), [Source 2](https://www.w3schools.com/jsref/met_element_addeventlistener.asp)).

Here is some background: I'm working on a Chrome Extension, and I decided to use Tippy.js. I found out this website I'm targeting, override `window.EventTarget.prototype.removeEventListener`, and they access `eventHandler.listener` without verifying it exists. I immediately assumed it was their mistake (to assume it has value), but after reading about it, I found out this information is mandatory.

atomiks:
[...]

PR #1028, opened 2022-01-30
Fix undefined listener invokes (bug #1027)

LazyCompiler:
Very simple solution for the bug I reported on:
[Tippy invokes removeEventListener and addEventListener with an undefined listener #1027
https://github.com/atomiks/tippyjs/issues/1027](https://github.com/atomiks/tippyjs/issues/1027)

Live demonstration available here: https://codepen.io/lazycompiler/pen/rNYxBjQ

Screen Shot 2022-01-31 at 1 20 17 AM

### DIFF
--- a/src/dom-utils.ts
+++ b/src/dom-utils.ts
@@ -107,8 +107,8 @@ export function updateTransitionEndListener(
   action: 'add' | 'remove',
   listener: (event: TransitionEvent) => void
 ): void {
-  if(listener) {
-      const method = `${action}EventListener` as
+  if (listener) {
+    const method = `${action}EventListener` as
       | 'addEventListener'
       | 'removeEventListener';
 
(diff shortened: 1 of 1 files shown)


src/dom-utils.ts
```diff
@@ -107,15 +107,17 @@ export function updateTransitionEndListener(
...
-  const method = `${action}EventListener` as
-    | 'addEventListener'
-    | 'removeEventListener';
-
-  // some browsers apparently support `transition` (unprefixed) but only fire
-  // `webkitTransitionEnd`...
-  ['transitionend', 'webkitTransitionEnd'].forEach((event) => {
-    box[method](event, listener as EventListener);
-  });
+  if(listener) {
```
atomiks, line 110: nit: there should be a space before the `(`. Did Prettier not run on commit?

LazyCompiler: Good point. Not sure why it happened. I initially used VS Code prettifier, but the style was too different, so I committed it the way it was.

atomiks: If you can update it this should be good to go

PR #1042, opened 2022-04-03
fix merging modifiers

ccpu:
In tippy, modifiers used in the code cannot be modified or overridden. I tried to modify the `preventOverflow` modifier, but it didn't work, so I added a few lines of code to merge the modifiers.

### DIFF
--- a/src/createTippy.ts
+++ b/src/createTippy.ts
@@ -677,7 +677,17 @@ export default function createTippy(
       });
     }
 
-    modifiers.push(...(popperOptions?.modifiers || []));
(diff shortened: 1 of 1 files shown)


atomiks:
It works fine though?: https://codepen.io/atomiks/pen/yvwQyZ

atomiks:
Note: Popper performs merging of modifiers for us, which is why it works. See: https://popper.js.org/docs/v2/faq/#how-do-i-set-modifier-defaults-and-allow-them-to-be-overridden

ccpu:
I do apologize, been using old version on my app, it does work indeed.

PR #1074, opened 2022-08-26
Update docs to show appendTo accepts an element

gamb:
Useful functionality that wasn't obvious from the docs


website/src/pages/v6/all-props.mdx
```diff
@@ -165,7 +165,7 @@ accessibility, clipping, or z-index issues.
 ```js
 tippy(targets, {
   // default (takes reference as an argument)
```
gamb, line 167: I see it is mentioned here but I think this can be made more explicit in the signature

PR #1166, opened 2024-04-19
delegate: Don't process event for non Element targets.

amanagr:
In Firefox, event targets can be pseudo elements like `text` nodes which don't have `.closest` method. This results in error when delegate tries to handle trigger for a text node target.


ojaha065:
We're a seeing this error a lot lately. Would be nice to have this merged and new version released. Sad to see this project abandoned

timabbott:
@atomiks any chance you can merge this PR? Even if you don't do a release, it'd be very helpful for us to be able to install a merged upstream commit for get rid of the spammy exception that this fixes.

